PR #2, opened 2016-04-26
add additonal bin reference

DavidWells:
You mean like this?

ref: https://github.com/kentcdodds/p-s/issues/1


kentcdodds:
Exactly. Thanks!

PR #4, opened 2016-05-02
feat(prefixes): Add improved prefixing

rowanoulton:
Hey, great project — I was intrigued by the mention of improved prefixing support in the roadmap and have made an attempt at providing that here.

I couldn't find anything to provide the kind of prefix resolution mentioned, so I put together a (somewhat experimental) package to do it [here](https://github.com/rowanoulton/prefix-matches). It's pretty crude, and I'm opening this pull request more to discuss the implementation early than to merge right in.

Let me know what you think. Would love to see this feature included.

Cheers! ✌️


kentcdodds:
This is fantastic! Thank you for working on it. Your package looks great! Thanks for making it a separate package to! That's the best way to solve this problem.

I'm sure you've noticed that the build broke because Node 0.10 and 0.12 don't support `const`. I recommend you either update your package source to support those old versions or create a build similar to the one for package-scripts.

Let me know if you need help!

rowanoulton:
Awesome. I did have a crack at fixing the issue, though I seem to be stuck between a rock and a hard place: removing `use strict` cause issues in the newer version of node, whereas keeping it causes issues in the older versions. How would you typically fix this? Avoid using `const`, `let`, etc altogether?

kentcdodds:
Yes. Either just use `var` or transpile your code

rowanoulton:
Alright cool, all fixed. I've made some tentative changes to the docs to reflect the deeper prefixing support this adds. Let me know if there's anything else I should add before merge.

ROADMAP.md
```diff
@@ -2,9 +2,9 @@
 
 ## Want to do
 
-- Add improved prefixing. Right now you can prefix the entire name of a script, but you couldn't prefix something like
-`test.watch` as `t.w` and I think that'd be sweet
 - Add the ability to specify multiple scripts that run in series (right now the solution is to just `&&`)
+- Add something like `p-s help` that will log out how to use the cli and available scripts to run in the project
```
kentcdodds: Do you think that you could [rebase your branch](http://kcd.im/gh-rebase)? These items should no longer be in the ROADMAP :-)

package.json
```diff
@@ -30,7 +30,6 @@
     "lodash.camelcase": "4.1.1",
     "lodash.contains": "2.4.3",
     "lodash.find": "4.3.0",
-    "lodash.get": "4.2.1",
```
kentcdodds: This is great! Thanks for updating this :+1:

kentcdodds:
It's almost perfect! Thanks!

rowanoulton:
Oops, that was a bad merge on my part. Fixed up and ready to go. Cheers! ✌️

kentcdodds:
Fantastic! Thank you so much! :D

kentcdodds:
Thanks again for the contribution @rowanoulton. I've added you as a committer. Feel free to continue filing pull requests as you feel the need :-)

rowanoulton:
Nice one. Cheers Kent!

PR #5, opened 2016-05-02
fix(casify): fix casify to also return original task name

abhishekisnot:
casify should also return original task name to avoid common user errors. For example, for task name
like e2e, it is not obvious to see that it converts to camelCase e2E and kebab-case e-2-e. This
avoids scripts must resolve to string error if user passes in the original task name over camelified
or kebabified task name

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1,31 +1,41 @@
 {
-  "projectName": "p-s",
-  "projectOwner": "kentcdodds",
-  "files": [
+  "projectName": "p-s"
+, "projectOwner": "kentcdodds"
+, "files": [
     "README.md"
-  ],
-  "imageSize": 100,
-  "commit": false,
-  "contributors": [
+  ]
+, "imageSize": 100
+, "commit": false
+, "contributors": [
     {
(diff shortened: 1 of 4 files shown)


kentcdodds:
Fantastic! Thank you!

kentcdodds:
Whoops, forgot to regenerate the contributors table. I'll do that a little later. Thanks.

kentcdodds:
Hey @abhishekisnot,
I just fixed the contributors stuff (we had an issue with the version). Turns out that GitHub wont size down your image. Could you upload an avatar to your GitHub profile that'll fix it I think:

screen shot 2016-05-02 at 9 59 15 am

abhishekisnot:
@kentcdodds Just updated github avatar. I think it looks good now.

jfmengels:
Ah, so the avatar url is not changed everytime the user avatar changes. Good to know!

kentcdodds:
Looks much better! Thanks!

kentcdodds:
Thanks again for the contribution @abhishekisnot. I've added you as a collaborator. Feel free to continue filing pull requests as you feel the need :-)

PR #9, opened 2016-05-18
feat: Allow user to specify envrionment variables for scripts

giladgo:
**What**:
This allows the user to specify env vars for a script or a set of scripts.

The great thing about it is that it allows to specify a set of env vars for multiple scripts at once. 
So, for example, specifying `env: { test: { foo: 'bar' } }` in the config will run `test.watch`, `test.mystuff` `test.otherstuff` with `foo=bar` as an env var.

**Why**:
Instead of doing this:
```js
scripts: {
  test: {
    bla: "NODE_ENV=TEST MODE=test_mode mocha bla/**/*.js",
    bli: "NODE_ENV=TEST MODE=test_mode mocha bli/**/*.js",
    blo: "NODE_ENV=TEST MODE=test_mode mocha blo/**/*.js",
  }  
}
```
You can do this:

```js
scripts: {
  test: {
    bla: "mocha bla/**/*.js",
    bli: "mocha bli/**/*.js",
    blo: "mocha blo/**/*.js",
  }  
}
env: {
  test: {
    NODE_ENV: "test"
    MODE: "test_mode"
  }
}
```
Much cleaner, isn't it? 😄 

**How**:
Code is pretty much self-explanatory. However if something is unclear, AMA!

### DIFF
--- a/bin/p-s.js
+++ b/bin/p-s.js
@@ -36,6 +36,7 @@ var psConfig = getPSConfig()
 
 runPackageScript({
   scriptConfig: getPSConfig().scripts,
+  envConfig: getPSConfig().env,
   scripts: scriptsAndArgs.scripts,
   args: scriptsAndArgs.args,
   options: merge(psConfig.options, {
--- a/bin/p-s.js
+++ b/bin/p-s.js
@@ -35,8 +35,8 @@ var scriptsAndArgs = getScriptsAndArgs(program)
 var psConfig = getPSConfig()
 
 runPackageScript({
-  scriptConfig: getPSConfig().scripts,
-  envConfig: getPSConfig().env,
+  scriptConfig: psConfig.scripts,
+  envConfig: psConfig.env,
   scripts: scriptsAndArgs.scripts,
   args: scriptsAndArgs.args,
   options: merge(psConfig.options, {
--- a/src/get-script-env.js
+++ b/src/get-script-env.js
@@ -1,0 +1,15 @@
+import prefixMatches from 'prefix-matches'
+import kebabAndCamelCasify from './kebab-and-camel-casify'
+
+export default function getScriptEnv(config, input) {
+  if (!config) {
+    return {}
+  }
+  config = kebabAndCamelCasify(config)
+  let env = prefixMatches(input, config)[0]
+  while (!env) {
+    input = input.split('.').slice(0, -1).join('.')
+    env = prefixMatches(input, config)[0]
+  }
+  return env
+}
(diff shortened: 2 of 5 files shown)


bin/p-s.js
```diff
@@ -36,6 +36,7 @@ var psConfig = getPSConfig()
 
 runPackageScript({
   scriptConfig: getPSConfig().scripts,
+  envConfig: getPSConfig().env,
```
kentcdodds: Could we optimize this by doing something more like:

```javascript
var psConfig = getPSConfig()
runPackageScript({
  scriptConfig: psConfig.scripts,
  envConfig: psConfig.env,
  // ...
})
```

Thanks!

kentcdodds:
Hmmm... I'm not certain about this feature. On the one hand you're right that it makes it much cleaner, but on the other I think it adds unnecessary complexity to our implementation.

Here's how I would solve your example above:

```javascript
const testEnv = 'NODE_ENV=TEST MODE=test_mode'
//
scripts: {
  test: {
    bla: `${testEnv} mocha bla/**/*.js`,
    bli: "${testEnv} mocha bli/**/*.js",
    blo: "${testEnv} mocha blo/**/*.js",
  }  
}
```

Much more straightforward and less ✨ magic.

I'm willing to be convinced. Curious what other repo watchers have to say. But I'd really prefer to keep things simple.

I realize you've worked hard on this and it's probably frustrating that I'm not super excited about it. I apologize for that. In the future, please file an issue for new features so we avoid this.

tleunen:
I also prefer your way, it's more obvious something will be done when running the command. There's no need to look for an `env` structure and make the link between the task names in scripts and `env`.

bin/p-s.js
giladgo: sure :)

giladgo:
I hear you.

@tleunen @kentcdodds all of your arguments are completely valid, and now I'm not 100% convinced 
about this feature either 😄 

But, let's examine the pros before we rule this one out:

 1. Even though this can be solved as @kentcdodds suggested (and in fact, that's how I solved it in one of my projects in the meantime :)), moving things to `env` will reduce clutter and improve readability even more. I agree that this adds to ✨  magic, but (as with everywhere in coding), it's a matter of trade-offs - magic v.s. simplicity/readability.
 2. No need to use the shell environment variable syntax. That is, no need to surround the values in quotes (and escape the quotes), etc. consider:

``js
build: "MY_NAME=\"Captain America\" npm build"
```
vs.
``js
build: "npm build"
// ..
env: {
  build: {
    MY_NAME: "Captain America"
  }
}
```

kentcdodds:
Yeah, after thinking of it further, I'm going to go ahead and close this. Forks are free though :-)

PR #10, opened 2016-05-18
docs(in-the-wild): Add "In the wild" section

nkbt:
**What**: Add "In the wild" section with links to 3rd-party libs using `p-s` as functional part

**Why**: Some usages that may not be initially intended as `p-s` goal, but can give inspiration and ideas for other users

**How**: -


kentcdodds:
LGTM!

Do you want to add yourself as a contributor? See the contributing guidelines. This would be considered an example contribution type

nkbt:
Why not :)

PR #12, opened 2016-05-19
docs(contributing): Fix the ROADMAP.md uri

tim-mcgee:
**What**: I fixed the link to the ROADMAP.md file in the CONTRIBUTING.md file

**Why**: Because the link wasn't working correctly

**How**: With a Text-Editor?

I started at a new feature similar to avas or tapes --require to specify a module which will be required before the actual config so the config could be written in es7 while I noticed the link was broken.


kentcdodds:
Sweet. One thing.

.all-contributorsrc
```diff
@@ -57,6 +57,17 @@
...
+    },
+    {
+      "login": "tim-mcgee",
+      "name": "Tim McGee",
+      "avatar_url": "https://avatars.githubusercontent.com/u/14267457?v=3",
+      "profile": "https://github.com/tim-mcgee",
+      "contributions": [
+        "code",
+        "doc",
+        "test"
```
kentcdodds: Looks like this contribution is only `doc`, could you remove `code` and `test`? (Or contribute code/tests) :-)

bin/p-s.js
```diff
@@ -46,7 +47,7 @@ runPackageScript({
 }, function(result) {
   if (result.error) {
     log.error(result.error)
-    const FAIL_CODE = 1
```
tim-mcgee: This file isn't compiled so it has to be Node 0.10.0 friendly. Const is only available behind a flag in Node 0.10.0 [node.green](http://node.green/#const)

bin/p-s.js
```diff
@@ -55,7 +56,22 @@ runPackageScript({
...
+        require(program.require)
+      } catch (e) {
+        log.warn({
+          message: colors.yellow('Unable to preload ' + program.require),
+          ref: 'preload-module'
+        })
+      }
+    }
+    var config = require(psConfigFilename)
+    if (programm.resolve === 'babel-register') {
```
tim-mcgee: The problem with babel-register is that you get the default export behind a module with a default property. This line fixes it but I think it's not really pretty :)

.all-contributorsrc
tim-mcgee: Just added code as well :)

bin/p-s.js
kentcdodds: Yikes! Thanks for noticing that! Would you like to [makeapullrequest.com](http://makeapullrequest.com)?

kentcdodds: :+1:

kentcdodds:
This LGTM. Could anyone else weigh in?

bin/p-s.js
```diff
@@ -55,7 +56,23 @@ runPackageScript({
...
+      } catch (e) {
+        log.warn({
+          message: colors.yellow('Unable to preload ' + program.require),
+          ref: 'preload-module'
+        })
+      }
+    }
+
+    var config = require(psConfigFilename)
+    if (program.require === 'babel-register') {
```
kentcdodds: Could we actually do:

```javascript
if (config.default) {
  config = config.default
}
```

That should be safer and allow people to use ES6 modules or CommonJS even if they do transpile with babel or anything else.

kentcdodds: Would probably be good to add a comment there as well.

kentcdodds:
Oh, and also... About tests.... Yeah, I'm not sure how to test this... We can leave it off for now.

bin/p-s.js
```diff
@@ -55,7 +56,23 @@ runPackageScript({
...
   try {
-    return require(psConfigFilename)
+    // Require a module like babel-register
+    if (program.require) {
+      try {
+        require(program.require)
+      } catch (e) {
+        log.warn({
+          message: colors.yellow('Unable to preload ' + program.require),
+          ref: 'preload-module'
```
kentcdodds: This `ref` should reference a line in [ERRORS_AND_WARNINGS.md](https://github.com/kentcdodds/p-s/blob/master/other/ERRORS_AND_WARNINGS.md). Similar to how [this](https://github.com/kentcdodds/p-s/blob/9ae2217d365c06153d47db2acfd63230fcff83fb/bin/p-s.js#L62) references [this](https://github.com/kentcdodds/p-s/blob/9ae2217d365c06153d47db2acfd63230fcff83fb/other/ERRORS_AND_WARNINGS.md#unable-to-find-config). If you could add a line to ERRORS_AND_WARNINGS.md, that'd be great 👍

kentcdodds:
I merged your other PR. You'll need to [rebase this branch](http://kcd.im/pr-rebase).

tim-mcgee:
I hope I did all things right :D

README.md
```diff
@@ -290,8 +295,9 @@ benefits of npm scripts I wanted to keep).
...
+
+-| :---: | :---: | :---: | :---: | :---: | :---: |
```
tim-mcgee: I failed while rebasing... I hope the list is fixed now.

kentcdodds: If you preview this, it looks like it's formatted incorrectly: https://github.com/tim-mcgee/p-s/blob/98aaff50503ec1811a6da755ab6fd7f0b085f361/README.md

Could you run `npm start generateContributors` to automatically update the table?

other/ERRORS_AND_WARNINGS.md
```diff
@@ -18,3 +18,11 @@ point to a file that actually exists.
 
 This happens if you specify a script name that can't be resolved to a script. See the docs/examples for how you can
 define scripts and how they can be referenced.
+
+## Unable to preload module
+
+This happens the module which you try to preload couldn't be resolved.
```
kentcdodds: This could probably more clearly call out the relationship with the `--require` flag. Something like:

> This happens when you use the `--require` flag and the module you specify cannot be resolved.

other/ERRORS_AND_WARNINGS.md
```diff
@@ -18,3 +18,11 @@ point to a file that actually exists.
...
+
+### To fix:
+
+Check if you spelled the module correctly.
```
kentcdodds: I would say:

> 1. Check that you spelled the module correctly
> 2. Check that the module you wish to require is require-able

bin/p-s.js
```diff
@@ -55,7 +56,25 @@ runPackageScript({
 function getPSConfig() {
   var psConfigFilename = program.config ? resolve(process.cwd(), program.config) : findUp.sync('package-scripts.js')
   try {
-    return require(psConfigFilename)
+    // Require a module like babel-register
+    if (program.require) {
+      try {
+        require(program.require)
```
kentcdodds: If someone wanted to `--require ./some-file/in/their-project.js` this wouldn't work. I think there's a way to use `resolve` to make this work relative to `process.cwd()`.

tim-mcgee: if (program.require.substr(0, 0) === '.') {
  program.require = path.resolve(process.cwd(), program.require)
}

kentcdodds: No, because they could do an absolute path or a relative path without a dot too

tim-mcgee: But if you require('some-file/in/your-project.js') node would cry too? So I think the only problem would be resolving the relative paths because I you don't need to resolve absolute paths because their already absolute and path.resolve('/some/absolute/path.js') would return '/some/absolute/path.js' anyway. Or I am wrong at this?

The other option I could imagine is Module.prototype.require.call(process.cwd(), programm.require) how about this?

kentcdodds: Good point. Yeah, let's just check if it starts with `.` 👍

kentcdodds:
I added you as a computer. Please feel free to continue to file PRs as you see fit. 😃

kentcdodds:
Also, looks like this branch has fallen behind again and needs to be rebased. Sorry!

tim-mcgee:
I added the `'.'` check and also added the options to the `src/bin-utils.js` file.
I tested absolute, relative and npm modules manually. Should work fine :)

kentcdodds:
This looks perfect! Thanks @tim-mcgee!

kentcdodds:
I'd like to get at least one other person to review. @rowanoulton, @abhishekisnot, or @giladgo, could one of you look at this and merge it?

bin/p-s.js
```diff
@@ -55,7 +56,29 @@ runPackageScript({
...
+          message: colors.yellow('Unable to preload ' + program.require),
+          ref: 'unable-to-preload-module'
+        })
+      }
+    }
+
+    var config = require(psConfigFilename)
+    // babel-register is loading the default package into a object with the prop
+    // default so if it exist use it as the config
+    if (config.default) {
```
tim-mcgee: Maybe we should check `if (config.__esModule === true)` instead of `if (config.default)`?

kentcdodds: Ah, yeah, that's what Babel does after all:

```js
'use strict';

var _x = require('x');

var _x2 = _interopRequireDefault(_x);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
```

tim-mcgee:
Maybe we should sometimes exchange all the underlying code so it works like babel so you could do named export as well?

````javascript
const scripts = {...}
const options = {...}

export {scripts, options} // hope this is the right syntax I am mainly using default exports...
````

**Comment by kentcdodds** [comment]
> Maybe we should sometimes change all the underlying code so it works like babel and you could do named exports as well?

Nah, let's keep it simple. Let's try to avoid multiple ways of doing the same thing.

File: bin/p-s.js [hunk 11]
```diff
@@ -55,7 +56,29 @@ runPackageScript({
 function getPSConfig() {
   var psConfigFilename = program.config ? resolve(process.cwd(), program.config) : findUp.sync('package-scripts.js')
   try {
-    return require(psConfigFilename)
+    // Require a module like babel-register
+    if (program.require) {
+      var requirePath = (program.require.substr(0, 1) === '.'
```
giladgo: isn't `program.require[0]` simpler?

giladgo:
Other than one small, comment, LGTM

kentcdodds:
Awesome! Thanks for working on this! :shipit:

PR #16, opened 2016-05-20
P-S Exits Now If No Config File Could Be Found

tim-mcgee:
**Why**:
P-S would terminate anyway because it couldn't find a script.


kentcdodds:
LGTM

PR #20, opened 2016-05-31
Fix nested tasks display in help

### LINKED ISSUE #19: Nested taks are not displayed properly in the help
Inside the help section, the task list may be wrong.

If I have a few nested tasks, like:
```
build: {
  x: {
    y: {
      description: "build X-Y",
      script: "do sth"
    }
  }
}
```
With this example, it only display `x.y - build X-Y - do sth`

tleunen:
Fix #19 

The prefix was incorrect for the nested tasks when displaying the help section. This PR will fix the issue.

The issue was cause because we only passed the current prefix when trying to get the nested script, therefore only the last 2 nested prefix were displayed instead of all prefixes.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -76,6 +76,17 @@
       "contributions": [
         "example"
       ]
+    },
+    {
+      "login": "tleunen",
+      "name": "Tommy",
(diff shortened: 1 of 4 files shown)


kentcdodds:
LGTM

kentcdodds:
Thanks for this!

PR #23, opened 2016-06-08
Don't swallow syntax errors in package-scripts.js

### LINKED ISSUE #22: syntax errors in package-scripts.js should not be swallowed.
- `p-s` version: 1.0.1
- `node` version: 6.2.1
- `npm` version: 3.9.3

Problem description:

If there is a syntax error in the `package-scripts.js` (in my case for example a missing comma) then it will fail to load. However your error message swallows the syntax error and just says that it cannot find the `package-scripts.js` file which was confusing for me.

Suggested solution:

Basically when [capturing the error here](https://github.com/kentcdodds/p-s/blob/adb01e92e111cf5a0d0c3ac4f42386bda04cc5bc/src/bin-utils.js#L86-L88)

Just add a check for syntax error:

``` javascript
if (e instanceof SyntaxError) {
   throw e;
}
```

I'll gladly make the change and submit a PR.

jisaacks:
**What**:

Fixes #22 

**Why**:

See Issue #22 

**How**:

Rethrows the caught error if it's a `SyntaxError`


.eslintignore
```diff
@@ -2,3 +2,4 @@ node_modules
 .nyc_output
 coverage
 dist
+test/fixtures/syntax-error-module.js
```
kentcdodds: Maybe instead we could add `/* eslint-disable */` to the top of that file.

kentcdodds: Ah, nevermind, I expect it will probably blow up before even parsing that comment right?

kentcdodds:
:shipit: :-) LGTM. Could you verify that we can't just add the comment instead? Other than that, this is great. Thank you!

kentcdodds:
Nevermind. I'm fine with this! Thanks a ton!

kentcdodds:
@jisaacks, I added you as a committer to the project. Please feel free to continue to file pull requests as you feel the need :-0

jisaacks:
@kentcdodds sounds good! Thanks.

PR #29, opened 2016-07-08
Fix running as child process

nkbt:
**What**:
When running as child process with arguments (like `--config smth`), `process.argv` is never empty, though commander has no args.

Also it would be logically more correct to check `program.args`, since later we use it as `program.args[0]`.

```sh
/Users/nkbt/nkbt/react-swap/node_modules/p-s/dist/bin-utils.js:90
    scripts = program.args[0].split(',');
                             ^

TypeError: Cannot read property 'split' of undefined
    at getScriptsAndArgs (/Users/nkbt/nkbt/react-swap/node_modules/p-s/dist/bin-utils.js:90:30)
    at loadAndRun (/Users/nkbt/nkbt/react-swap/node_modules/p-s/dist/bin/p-s.js:47:56)
    at Object.<anonymous> (/Users/nkbt/nkbt/react-swap/node_modules/p-s/dist/bin/p-s.js:43:3)
 ```

**Why**

I want to proxy p-s from my component like

```js
require('child_process')
  .spawn(ps, ['--config', config].concat(process.argv.slice(2)), {
    cwd: process.cwd(),
    env: process.env,
    stdio: [process.stdin, process.stdout, process.stderr]
  });
```

To avoid adding extra dependency.

### DIFF
--- a/src/bin/p-s.js
+++ b/src/bin/p-s.js
@@ -19,7 +19,7 @@ program
   .on('--help', onHelp)
   .parse(process.argv)
 
-if (process.argv.length < 3) {
+if (program.args.length < 1) {
   program.outputHelp()
 } else {
   loadAndRun()


kentcdodds:
This makes sense and as long as the build passes I say :shipit: :-)

nkbt:
Cheers 🍻

kentcdodds:
FYI, this broke the following case:

```
p-s -p lint,test
```

Because `program.args` is an empty array in this case. I'm working on a fix now.

kentcdodds:
Also @nkbt, as a thanks for the contribution, I've given you write access to this repo. Please continue to make pull requests as you see fit :-)

nkbt:
Oh wow, sorry for the break, I don't have this case =(... Prob we need some e2e tests

kentcdodds:
I should set up `dont-break` with this project on travis. Do you have any suggestions for projects I can do that with?

kentcdodds:
The problem I see with using `dont-break` is that it could be that the `p-s` project is working fine, but the dependency's tests are broken. I think just regular e2e tests would be good. Maybe we could use [`nixt`](https://github.com/vesln/nixt).

nkbt:
I would go for just a shell script for CI/local =). Bash for the win. something like

```sh
npm start blah > test.log
cat test.log | grep "all good"
```

I have bunch of this sort of e2e tests for our cli tools and docker images.

kentcdodds:
Created #31 for anyone to take ownership 👍

nkbt:
I'll get some basic stuff done since I have bunch of little scripts ready
to use.

nkbt:
Nixt looks pretty sick by the way. Ill give it a try here

kentcdodds:
Sweet! 👍

PR #32, opened 2016-07-10
Use exact dependency versions and add npmrc with save-exact setting

nkbt:
**What**:
- [x] fix dependency to be exact version
- [x] add `.npmrc` config with save-exact so next time anyone runs `--save` or `--save-dev` - version will be saved without `^`

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+save-exact = true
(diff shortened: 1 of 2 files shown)


kentcdodds:
LGTM. Could you make sure that your pull requests follow the commit message mentioned in the contributing guidelines? Thanks!

nkbt:
Sure, sorry

PR #34, opened 2016-07-14
[wip] Add End-2-End tests for CLI

### LINKED ISSUE #31: Write e2e tests
As suggested by @nkbt in #30. Maybe we could use [`nixt`](https://github.com/vesln/nixt).

nkbt:
Fixes #31 

**What**:
- [x] Added `cli-tester`
- [x] Windows/OSX compatible tests
- [x] Add test for running `p-s` without any arguments
- [x] Add test for running with config file but without any arguments
- [ ] More tests

**Why**:
Make sure that expected behavior is not broken with different updates

This is WIP, so later will rewrite commit messages, fix code style, etc.

Some of used functions may be abstracted, though I think it is ok to keep them with tests, closer to context.

Would be great to have some more common usecases to test. Maybe another 2-3 to play with. Then anyone could be able to add more tests when new functionality is added.

Feel free to object any of decisions made, let me know what looks good, what is not so.

Cheers!

### DIFF
--- a/src/bin/p-s.test.js
+++ b/src/bin/p-s.test.js
@@ -1,28 +1,26 @@
+/* eslint func-style:0, max-len:[2, 130] */
 import path from 'path'
+import fs from 'fs'
+import os from 'os'
+import {createHash} from 'crypto'
 import test from 'ava'
-import tester from 'cli-tester';
-import fs from 'fs';
-import os from 'os';
-import {createHash} from 'crypto';
+import tester from 'cli-tester/es5'
 
+const babel = require.resolve('babel-cli/bin/babel-node')
 
-const babel = require.resolve('babel-cli/bin/babel-node');
-const cli = require.resolve('./p-s');
-
-
-const hash = string => createHash('md5').update(string).digest('hex');
-const tmpConfig = string => path.join(os.tmpdir(), `tmp-config-${hash(string)}.js`);
-const writeConfig = (file, string) => new Promise(resolve => fs.writeFile(file, string, resolve));
-const rmConfig = (file) => new Promise(resolve => fs.unlink(file, resolve));
+const hash = string => createHash('md5').update(string).digest('hex')
+const tmpConfig = string => path.join(os.tmpdir(), `tmp-config-${hash(string)}.js`)
+const writeConfig = (file, string) => new Promise(resolve => fs.writeFile(file, string, resolve))
+const rmConfig = file => new Promise(resolve => fs.unlink(file, resolve))
 
 
 test('p-s: run without args', t =>
   tester(babel, require.resolve('./p-s'))
     .then(({code, stdout, stderr}) => {
-      t.is(code, 0, 'should exit with code 0');
-      t.regex(stdout, /Usage: p-s \[options]/, 'should show help by default');
-      t.is(stderr, '', 'should not have any errors');
-    }));
+      t.is(code, 0, 'should exit with code 0')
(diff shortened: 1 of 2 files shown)


src/bin/p-s.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+
+const hash = string => createHash('md5').update(string).digest('hex');
+const tmpConfig = string => path.join(os.tmpdir(), `tmp-config-${hash(string)}.js`);
+const writeConfig = (file, string) => new Promise(resolve => fs.writeFile(file, string, resolve));
+const rmConfig = (file) => new Promise(resolve => fs.unlink(file, resolve));
+
+
+test('p-s: run without args', t =>
+  tester(babel, require.resolve('./p-s'))
```
nkbt: Kind of weird way to babelify scripts... I thought cli-tester could be updated, but not sure how exactly update it to avoid any extra complexity.

Same for config files. This looks like a nice thing to abstract. Maybe I will add these as just helpers that one can use.

src/bin/p-s.test.js
```diff
@@ -0,0 +1,53 @@
...
+import os from 'os';
+import {createHash} from 'crypto';
+
+
+const babel = require.resolve('babel-cli/bin/babel-node');
+const cli = require.resolve('./p-s');
+
+
+const hash = string => createHash('md5').update(string).digest('hex');
+const tmpConfig = string => path.join(os.tmpdir(), `tmp-config-${hash(string)}.js`);
```
nkbt: Since tests are run in parallel, I need each file to be unique per test.

src/bin/p-s.test.js
```diff
@@ -0,0 +1,53 @@
...
+const writeConfig = (file, string) => new Promise(resolve => fs.writeFile(file, string, resolve));
```
nkbt: actually writing real file is the only way to go. Mock-fs/memory-fs do not work, since testing is done in a separate process.

kentcdodds:
So far I really like it!

boneskull:
This should probably be merged if the code is ok (even if incomplete).  It would have likely avoided releasing v2.4.0 and v2.4.1 :wink:

kentcdodds:
True. Need to get the Travis stuff to work though.

kentcdodds:
Looks like it's just some linting issue on Node 6. With the older node stuff it's complaining about ES6. I think we'd probably be ok with just running this on Node 6 though, so maybe we can make that work.

kentcdodds:
Would be easier to do this if [this](https://twitter.com/kentcdodds/status/770296988382343168) existed.

kentcdodds:
Actually, that cli wouldn't help because this is working as a normal test (forgot about that).... Hmmm... I guess that `cli-tester` does export an es5 version. Should solve some of the problems if we use that.

boneskull:
@kentcdodds @bahmutov was working on [something like this](https://glebbahmutov.com/blog//precompiled-javascript/)

kentcdodds:
Alrighty, I think that I've got it working. Let's see what Travis thinks of this.

kentcdodds:
LGTM

kentcdodds:
Awesome. Thanks for the ping @boneskull. Thanks for the work @nkbt! If anyone's bored, we could probably use some more tests here :)

nkbt:
Thanks.

Sorry I haven't completely done what I planned... Just buried under my current work-related projects, so my oss stuff is on hold for a moment

kentcdodds:
You're good buddy :) Thanks for your initial work on this!

Issue #33, opened 2016-07-12
Default root script


tleunen:
Would you be willing to allow a default root script in order to be able to run the script with just `npm start`, or `p-s`?

```js
module.exports = {
    scripts: {
        default: {
            script: 'echo default script'
        }
    }
}
```

kentcdodds:
I could see that being valuable. In that scenario, how do you propose to show the `help`?

kentcdodds:
Yeah, I agree that this would be a good thing to add. Let's do it as you've said :+1:

1. If there is no `default` script, then show help
2. If there's a `default` script, then run that (don't show help)
3. If someone provides `--help` (only) then don't run any scripts and just show help (this means that `npm start foo -- --help` would not log help and run `foo`).

Sound good?

kentcdodds:
Because this is a breaking change, whoever makes the PR for this feature, please make it to the [`next` branch](https://github.com/kentcdodds/p-s/tree/next) so we can get a few breaking changes in a single release.

nkbt:
Good idea

PR #36, opened 2016-07-26
add init command

### LINKED ISSUE #35: Add init script
It'd be awesome to have an init script that read the `package.json` scripts and transferred them to a `package-scripts.js` and update the `scripts` in `package.json` to use the recommended `start` script.

kentcdodds:
Closes #35


kentcdodds:
I'd love it if someone could give this a review. Notice that I'm merging into `next`. I want to implement #33 before this gets released.

As far as I'm concerned though this LGTM :)

PR #37, opened 2016-07-26
feat(cli): add support for root default script

### LINKED ISSUE #33: Default root script

kentcdodds:
**What**:

Adds support for the use case where running `p-s` without arguments will run `scripts.default`

**Why**:

Because often servers are configured to run `npm start` by default.

**How**:

- check for any arguments and default to `['default']` if no scripts are given.
- Update the `init` script to account for a `start` script

Closes #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Put all of your scripts in a file called `package-scripts.js` and use `p-s` in a
 ```javascript
 module.exports = {
   scripts: {
+    default: 'node index.js',
     lint: 'eslint .',
     test: {
       default: 'ava',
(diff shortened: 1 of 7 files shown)


kentcdodds:
Note, this is intended to be merged after the `init` PR (at which time the diff will be a bit cleaner). Also, this is intended to first be merged into the `next` branch.

kentcdodds:
LGTM

tleunen:
Great. Thanks @kentcdodds !

PR #38, opened 2016-07-26
feat(PATH): add npm bin to path

### LINKED ISSUE #27: feat request: Explicitly add node_modules/.bin to PATH
See [this comment](https://github.com/kentcdodds/p-s/issues/26#issuecomment-229389943). I could see value in explicitly adding `node_modules/.bin` to PATH so this behaves more like normal npm scripts.

I'm open to suggestions/ideas/alternative opinions.

kentcdodds:
**What**:

This will search for a local node_modules/.bin directory and attempt to add it to the PATH (in a cross-platform way using `manage-path`).

**Why**:

So people can use these just as they would normal npm scripts. The only case where the PATH doesn't actually contain the npm bin is when you run the `p-s` script directly (rather than using npm scripts). This is an edge case, but it's still handy

**How**:

Use `find-up` to find the `node_modules/.bin` and `mange-path` to add it to the PATH in a cross-platform way

Closes #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,8 +204,12 @@ module.exports = {
     simple: 'echo "this is easy"', // npm start simple
     test: {
       default: {
-        script: 'echo "test things!"', // npm start test
-        description: 'just pass it on to npm... Do not take this config very seriously :-)',
+        script: 'ava', // npm start test
+        description: 'Run tests with ava',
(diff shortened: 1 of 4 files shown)


kentcdodds:
LGTM

This can be merged into `next` anytime.

PR #39, opened 2016-07-26
fix(default): Now default scripts will be a fallback only

### LINKED ISSUE #25: Don't include `default` as one of the prefixable scripts
Given these scripts:

```javascript
module.exports = {
  scripts: {
    thing: {
      default: 'echo "default"',
      dee: 'echo "dee"', 
    },
  },
}
```

If I run `npm start thing` it will run `thing.default` as expected/documented. Same if I run `npm start t` which is great. But if I want to run the `thing.dee` script, the best I can get with the prefix resolver is `npm start t.dee` because running `npm start t.de` will run the default script.

So I propose that we make a (breaking) change to exclude `default` from the prefix resolver. @rowanoulton, would you be interested in taking this?

Because this and #24 are both breaking changes, we should probably merge them into a branch called `next` and when they're both merged into that, we'll merge that into the `master` branch so we can release them at the same time.

kentcdodds:
**What**:

This adds logic to check first for non-default scripts, then fallback on
default scripts if none are found that satisfies the input

**Why**:

Because #25

**How**:

Create a clone of the config object and remove all default scripts,
check the prefix with that, if none is found attempt again with the
original config object.

Closes #25

BREAKING CHANGE: You're unlikely to have this problem, but if you had a
script that matched the prefix `def` for example, the default script
would be run. Now, your specific script will be run instead.

### DIFF
--- a/src/get-script-to-run.js
+++ b/src/get-script-to-run.js
@@ -1,3 +1,4 @@
+import {each, cloneDeep, isPlainObject} from 'lodash'
 import prefixMatches from 'prefix-matches'
 import resolveScriptObjectToString from './resolve-script-object-to-string'
 import kebabAndCamelCasify from './kebab-and-camel-casify'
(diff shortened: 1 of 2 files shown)


kentcdodds:
LGMT

kentcdodds:
LGTM

Issue #48, opened 2016-08-22
Should p-s stop executing commands when an error occurs?


tleunen:
I wonder if p-s should stop the execution when an error occurs when executing multiple commands (in parallel or not).

Any thoughts?

Especially when p-s run a command and then the second command depends on the first one...

kentcdodds:
Yes it should. I thought about this just recently. I'm happy to accept a pull request to do this!

kentcdodds:
I think I'd like to get [this](https://github.com/mysticatea/npm-run-all/issues/59) done before working on this.

PR #53, opened 2016-09-03
abort all processes when one fails

kentcdodds:
ref #48


src/index.js
```diff
@@ -18,40 +19,106 @@ function runPackageScripts({scriptConfig, scripts, args, options = {}}, callback
...
   }
   const scriptNames = arrify(scripts)
-  const method = options.parallel ? 'map' : 'mapSeries'
-  async[method](scriptNames, (scriptName, cb) => {
-    const child = runPackageScript({scriptConfig, options, scriptName, args})
-    if (child.on) {
-      child.on('exit', exitCode => cb(null, exitCode))
-    } else {
-      cb(child)
+  if (options.parallel) {
```
kentcdodds: I feel like the code below could be greatly simplified with RxJS. But I have no idea how to use Observables. Someone please help me!

SamVerschueren: I don't really see an opportunity for observables. Might be missing something though. Why do you think observables make sense here?

kentcdodds: I've never used Observables, but I thought that they're supposed to really help with complex async stuff. :shrug:

SamVerschueren: I think it can be done a little bit cleaner though. I'll  do a PR later today or this weekend.

ta2edchimp: Maybe [`zip`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/zip.md) is what you're looking for (re RxJS)? Merge multiple Observables/Promises, subscribe to the result and have ha single handling of their results ...

kentcdodds: I don't know what I'm looking for.. I just think this code could be cleaned up a bit...

SamVerschueren: When running in parallel, you can just use `Promise.all`, when executing serially you can do something like this.

```js
const result = promises.reduce((res, task) => {
   return res.then(() => task());
});

result.then(...);
```

kentcdodds:
Sweet I think this is good to :shipit: 

Could I get some people to test it? I've released it as the beta: `npm install p-s@beta`. Just make sure your stuff's not broken?

cc @boneskull, @tleunen, @nkbt, @DavidWells, @Hypercubed, @rowanoulton 

This is a significant refactor and I'd hate to break anyone due to this bug fix! (Ref #48)

kentcdodds:
:shipit:

kentcdodds:
This is good to merge by anyone as soon as it's reviewed!

src/index.js
```diff
@@ -1,57 +1,125 @@
-import spawn from 'spawn-command'
-import async from 'async'
+import spawn from 'spawn-command-with-kill'
+import Promise from 'bluebird'
 import colors from 'colors/safe'
-import {isString, find, clone} from 'lodash'
+import {isString, clone} from 'lodash'
```
SamVerschueren: Full lodash is quite huge, why not install them separately?

Hypercubed:
I gave it a quick run on OSX and windows.  No issues other than an unrelated issue on windows (where you aware that the command `package-scripts` fails on windows?).

So... LGTM.

kentcdodds:
> where you aware that the command `package-scripts` fails on windows?

Oh no! I wasn't. Could you file an issue for that? Thanks for testing it out @Hypercubed!

kentcdodds:
:shipit:

src/index.js
kentcdodds: I get you, but honestly I'm using about a dozen methods and I'd rather avoid installing each separately/managing them. It's 1.4MB (uncompressed) and the likelihood that someone's already depending on lodash in their project (even if it's transitively) is quite high.

Thanks for the suggestion.

boneskull:
haha, I need more than 22h notice :wink:

kentcdodds:
I guess I was anxious and excited :-)

boneskull:
Have you considered therapy? :wink:

PR #55, opened 2016-09-05
rename to `nps`

kentcdodds:
**What**: Renames the package to `nps`.

**Why**:
- Windows issues #54
- `p-s` is hard to type (so there's an `nps` alias anyway)
- It's confusing to have several aliases

**How**:
- Find/replace mostly. Then I did some magic with the scripts to make things work without a locally installed instance of `nps` (because it doesn't exist yet).

Another snag is the `nps` package is actually already taken (I think this is actually why I named the package `p-s` in the first place). So [I've requested to get ownership of `nps`](https://github.com/tmpvar/nps/issues/2). We'll see what the owner says. If the request is rejected, then I think I'll use `npm-package-scripts` as the package name. Not as terse, but that's ok.

Remaining Steps:

- [ ] Determine package name (`nps` or `npm-package-scripts`)
- [ ] Rename repo
- [ ] Update Travis/DependencyCI/etc.
- [ ] Create a new screencast
- [ ] Deprecate old package (in a breaking change release because let's face it, who wants to see deprecation warnings every time they install stuff that still works)

I'd really appreciate feedback/review on this. Thoughts? 💭


README.md
```diff
@@ -32,19 +32,19 @@ which has fundamental issues (like no comments).
...
 {
   "scripts": {
-    "start": "package-scripts"
+    "start": "npm-package-scripts"
   }
 }
 ```
 
-**package-scripts.js**
+**npsfile.js**
```
sxn: To me, `npsfile` sounds a bit nondescript, since `file` in and of itself doesn't say anything about what function it's supposed to serve. Maybe `config` or something to that effect would be clearer?

kentcdodds: Good point :+1:

What if we simplify things and just have it be: `package-scripts.config.js`

That would make life MUCH easier and we wouldn't need to rename the project either :)

sxn: :+1: for `package-scripts.config.js`

I'm on the fence about the renaming of the project. Typing `p-s` is obviously more cumbersome than `nps`, so I'd be very much in favor of having it as an alias. However I'm not sure if I would rename the whole project to `nps` -- seeing it causes my brain to immediately process it as `npm`. On top of that, it would also take time to see if the maintainer of the current `nps` is willing to part with the name - plausible since the project seems to have moved from the CLI to http://npmsearch.com/ . If they agree to part with the name, they might still choose to keep their project on the registry. In that case, I actually think the logistics of having two tools inhabit the same 'space' would be too painful to deal with for it to be a good option.

`p-s` doesn't have any of those problems. Sure, it's more difficult to type than `nps`, but the alias can still stay there. Would there still be three aliases?

I guess what I'm getting at is that, if the project is to be renamed, I would probably rename it to something that more clearly states what it is rather than something that, best case scenario, doesn't confuse things any more than they currently are.

kentcdodds: Sounds good to me! I'll continue to use the `nps` alias. I'll also make another pull request to change the default config filename. This will be a breaking change so it'll be a major version bump. Anything else major you think should change?

nkbt:
You need to do something similar to what Michael Jackson did with npmcdn, open PRs for all dependent repos on github =)

PR #56, opened 2016-09-06
feat(cli): allow hiding scripts from the help list

sxn:
**What**: Allow scripts to be hidden from the help list.

**Why**: In order to allow having semi-private scripts - semi because they will still be runnable as long as you know they're there.

**How**: by changing the ```src/resolve-script-object-to-string.js::resolveScriptObjectToScript``` function so that if the ```hiddenFromHelp``` property is defined and truthy, undefined is returned.


src/resolve-script-object-to-string.test.js
```diff
@@ -32,7 +44,7 @@ test('resolves default to the script if it is a string', t => {
   t.is(result, 'string')
 })
 
-test('does not resolve a script object without a script or defualt', t => {
+test('does not resolve a script object without a script or default', t => {
```
kentcdodds: 👍

src/resolve-script-object-to-string.js
```diff
@@ -12,13 +12,22 @@ function resolveScriptObjectToString(script) {
...
-    }
+    return resolvePlainObjectToScript(script)
   } else if (isString(script)) {
     return {script}
   }
   return undefined
 }
+
+function resolvePlainObjectToScript(script) {
+  if (!isUndefined(script.hiddenFromHelp)) {
```
kentcdodds: I think this should probably simply be: `if (script.hiddenFromHelp) {`

I think we're good to go with `truthy`. So `hiddenFromHelp: false` should not hide it from help ;-)

sxn: Huh. I could've sworn that I pushed the truthy-checking version. Obviously I didn't though. :fearful:

kentcdodds:
LGTM

kentcdodds:
@sxn, thanks so much for this feature! I've added you as a collaborator on the project. Accept that invitation and feel free to squash and merge this PR yourself :)

Feel free to watch the repo and continue making contributions as you feel the need/desire :) Thanks again!

sxn:
Nice! Thanks, will do. :)

kentcdodds:
Oh one sec! Could you add some docs in the README for this feature?

sxn:
Sure thing!

kentcdodds:
LGTM

PR #57, opened 2016-09-07
chore(tests): migrate AVA to Jest

kentcdodds:
**What**: Migrate our tests from `ava` to `Jest`.

**Why**: It has some reeeeally awesome features. Especially snapshot testing which makes testing our CLI much simpler.

**How**: It wasn't easy. And we're not totally done. There's a bit of work yet to be done with getting coverage up. There are 2 tests I couldn't get working.


src/resolve-script-object-to-string.test.js
```diff
@@ -1,50 +1,49 @@
...
   const command = resolveScriptObjectToString({default: {script: buildCommand}})
-  t.is(command, buildCommand)
+  expect(command).toBe(buildCommand)
 })
 
-test('returns undefined if the object with default cannot be resolved to a string', t => {
+test('returns undefined if the object with default cannot be resolved to a string', () => {
   const result = resolveScriptObjectToString({default: {blah: 'stuff'}})
-  t.is(result, undefined)
+  expect(result).toBe(undefined)
```
cpojer: It's recommended to do `toBeUndefined()` to get better failure messages.

kentcdodds: Cool, thanks!

src/bin-utils/index.test.js
```diff
@@ -1,114 +1,113 @@
...
+  expect(scripts.length).toBe(0)
+  expect(args).toBe('')
 })
 
-test('preloadModule: resolves a relative path', t => {
-  const relativePath = './fixtures/my-module'
+test('preloadModule: resolves a relative path', () => {
+  // this is relative to process.cwd() I think...
+  // Because of some fancy stuff that Jest does with requires...
+  const relativePath = './src/bin-utils/fixtures/my-module'
```
kentcdodds: @cpojer could you explain why I have to do this? [Here's the relevant code](https://github.com/kentcdodds/p-s/blob/master/src/bin-utils/index.js#L105-L112) from `preloadModule`.

cpojer: `preloadModule` has its own copy of require so everything you require inside of `preloadModule` needs to be relative to `preloadModule`. I'd recommend to pass an absolute path to `preloadModule` in your tests by doing.

`path.resolve(__dirname, relativePath)` or something similar.

kentcdodds: Makes sense. Except part of what I'm testing is how that function deals with absolute/relative/node_modules paths.

kentcdodds: Actually.. it doesn't make sense. The paths I passed before were relative to `preloadModule`. But with Jest, they now have to be relative to `process.cwd()`

cpojer: https://github.com/kentcdodds/p-s/blob/master/src/bin-utils/index.js#L83

if you care about what process.cwd gives you, I'd recommend doing `process.cwd = jest.fn(() => '/some/path')`. Your test shouldn't rely on the real value of process.cwd.

kentcdodds: Aaaaahhhhhhh, now that makes sense. Thanks for pointing me to my own code. Totally forgot about that bit :-) thanks!

kentcdodds:
Could I get at least one other person to review these changes please? I think we're good to :shipit:

nkbt:
I'd be happy to, but I am tape guy and both ava and jest look too complicated for me 😜

cpojer:
If the tests are passing, shipit? :)

kentcdodds:
Looks like I've gotta fix a snapshot that's depending on my local environment. No trouble there.

I try to make it so this isn't just my project. Anyone else care?

cpojer:
Node 6.5 is inferring function names so if you use them as part of a snapshot, the snapshot might differ between 6.5 and previous versions. This is kinda unfortunate.

kentcdodds:
Fun Fact: Jest is only supported on Node v4. I'm cool with that :)

kentcdodds:
I've removed Node v0.12 and v0.10 from Travis. I think we technically still support these versions, but I don't think we need to run those on CI anymore.

kentcdodds:
🚢

PR #58, opened 2016-09-10
chore(cli): remove package-scripts alias

sxn:
**What**: remove `package-scripts` as an alias

**Why**: see #54 

**How**: just removing it from `package.json` and updating the README & tests.

This removes the 'package-scripts' alias as it conflicts with 'package-scripts.js' on Windows

issue: #54
breaking: 'package-scripts' won't be available as an alias anymore


kentcdodds:
Looks great! What do you think about using `nps` rather than `p-s`? I like the `nps` alias better and want to encourage others to use it more.

sxn:
The less I have to move my hands on the keyboard, the better, so I use `nps` too :smile:

kentcdodds:
Yeah, let's have most of the documentation use `nps`. Maybe we can add a note about why at the beginning somewhere.

sxn:
Done :) I left occurrences of `p-s` when they were referring to the package itself not the CLI, but I might have missed some/also changed some cases that I shouldn't've. Can someone have a look please?

README.md
```diff
@@ -32,14 +32,14 @@ which has fundamental issues (like no comments).
 
 ## This solution
 
-Put all of your scripts in a file called `package-scripts.js` and use `p-s` in a single `package.json` script:
+Put all of your scripts in a file called `package-scripts.js` and use `nps` in a single `package.json` script:
```
kentcdodds: Let's mention that `nps` is the recommended alias right here. Otherwise people may be confused.

README.md
```diff
@@ -32,14 +32,18 @@ which has fundamental issues (like no comments).
...
+
+Note: We used `nps` even though the package we just installed is named `p-s`. This is because internally we use `nps`
```
kentcdodds: Could we note that `nps` is an acronym for `npm package scripts`?

kentcdodds:
LGTM

kentcdodds:
TIL why it's important to keep PRs up-to-date with `master` before merging them :)

screen shot 2016-09-12 at 11 48 07 am

sxn:
Thanks! Sorry for not answering, I'm currently on holiday so I'm only near my computer in the morning or evening, and coupled with the timezone difference it can get .. odd. :crying_cat_face:

kentcdodds:
No worries!

PR #60, opened 2016-09-12
docs(README): update title to use `nps`

kentcdodds:
I think this will reinforce the use of the `nps` alias. But worry that it'll make things a bit more confusing because `p-s` is the name of the repo/package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
-# npm-package-scripts (aka nps)
+# p-s
+
+> aka npm-package-scripts or `nps` for short
 
 All the benefits of npm scripts without the cost of a bloated package.json and limits of json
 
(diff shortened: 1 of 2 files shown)


kentcdodds:
LGTM

kentcdodds:
I plan on making another video eventually that changes how I refer to `p-s` (to be `nps`). Could I get someone to review these changes? You can read it [here](https://github.com/kentcdodds/p-s/blob/103328c09b7f0ea88094e9073061d5476346b585/README.md)

sxn:
is this intended to be this way? Other than that it LGTM

kentcdodds:
Yes, it's confusing, but I want people using `nps` and it'd be challenging to make that message change based on the script run. Unless we were using yargs (which I want to do eventually)

Issue #63, opened 2016-09-26
Support YAML config


bfred-it:
It'd be nice if p-s supported YAML configuration alongside the JS one

CommonJS files are slightly easier to read/write than JSON, but JS objects are still full `'{"\,}`

```js
module.exports = {
  scripts: {
    default: 'node index.js',
    lint: 'eslint .',
    test: {
      // learn more about Jest here: https://kcd.im/egghead-jest
      default: 'jest',
      watch: {
        script: 'jest --watch',
        description: 'run in the amazingly intelligent Jest watch mode'
      }
    },
    build: {
      default: 'webpack',
      prod: 'webpack -p',
    },
    validate: 'nps --parallel lint,test,build',
  },
}
```

in YAML it would just be

```yml
default: node index.js
lint: eslint .
test:
  # learn more about Jest here: https://kcd.im/egghead-jest
  default: jest
  watch:
    script: jest --watch
    description: run in the amazingly intelligent Jest watch mode
build:
  default: webpack
  prod: webpack -p
validate: nps --parallel lint,test,build
```

Plus, even the syntax highlighter is nicer to YAML :P

YAML isn't dynamic like JS, but probably most people don't use that (and if they do, they can just use the standard CJS syntax)

kentcdodds:
I'd be ok with that. Mind filing a PR?

bfred-it:
I can try, but not this week

kentcdodds:
Sounds good to me. Let me know if you need a hand.

sxn:
I may have time to look into it this week too, if needed. 😄

kentcdodds:
I'm sure that @bfred-it would appreciate it :) It's unlikely that I'll work on this.

sxn:
:shipit:

bfred-it:
Looks great!

I was wonder though why that extra `script:` level was necessary, that's why I didn't include in my example

kentcdodds:
Yeah, it's so you can specify `options` :)

bfred-it:
Duh! Didn't even notice that. Thanks!

sxn:
Yeah, sorry for doing that without asking! Initially I implemented it _without_ the scripts level, but that felt a bit off.

As a side note: would it make sense to make it so that the `init` script can create a `package-scripts.yml` file too?

kentcdodds:
Oh yeah, great idea! Feel free to file an issue for that :+1:

PR #64, opened 2016-09-29
feat(cli): add .yml config support

sxn:
**What**: allowing users to configure `nps` using a YAML config file instead of JS

**Why**: to improve readability

**How**: by leveraging https://github.com/nodeca/js-yaml

Wasn't sure how much I should touch the `src/bin-utils/index.js` file so I decided to do it as little as possible 🐱 

Allow configuring nps in `package-scripts.yml` instead of `package-scripts.js`

issue: #63


sxn:
Hm. Seems I missed something when running the tests.

sxn:
Ha. One of the problems seems to be that the `cli-test` is looking for `package-scripts.yml` inside its own folder, doesn't find it so it proceeds in searching for it in the parent folder, where it *does* find it but, of course, it's not the config it was looking for. Should we do something to disallow this? (apart from actually adding the file, of course)

package-scripts.yml
```diff
@@ -0,0 +1,48 @@
+scripts:
```
kentcdodds: Interesting. Does this PR add a `package-scripts.yml` in addition to the existing `package-scripts.js` file?

src/bin/nps.js
```diff
@@ -82,6 +82,10 @@ function getPSConfig() {
   return config
 }
 
+function getPSConfigFilepath() {
+  return program.config || findUp.sync('package-scripts.yml') || findUp.sync('package-scripts.js')
```
kentcdodds: Hmmm... I'm thinking that more people will use the `.js` than those who use the `.yml`. Could you put the `.js` first here?

sxn: Sure!

package-scripts.yml
sxn: It seems we understood issue #63 differently. What was your idea of how it would work? It shouldn't be a problem to change it so that we're all on the same page 😃

kentcdodds: Haha, I didn't think we'd need to have this project use a yml file. The js version should be sufficient.

sxn: Hm. So we'd just have `package-scripts.js`, but it could either have YAML or JS inside, and the script should figure out what's what?

kentcdodds: Hmm... Maybe I'm confused... But that's not what I meant. I'm just not sure why you're adding a package-scripts.yml file to this repo.

sxn: Ah, yeah. No, there's no real reason for that to be here, in fact.

sxn:
Cool! Just:
lgtm

😄

kentcdodds:
:shipit:

sxn:
I seem to have broken [dependencyci](https://dependencyci.com/github/kentcdodds/p-s/builds/105)? What am I doing wrong?

kentcdodds:
I don't think you're doing anything wrong. I think something's wrong with dependencyci. We can probably just wait I'll rerun it again later. Thanks!

PR #68, opened 2016-10-04
feat(cli): allow initializing a YAML config

### LINKED ISSUE #66: Allow initializing to a YAML file
Currently, users who want to start using `p-s` in an existing project can hit the ground running by using `nps init`. This will generate a `package-scripts.js` file from their `package.json` file and update the latter to utilize the `nps` binary.

Since we now have YAML config support (#64), it would be great if `nps init` would allow generating a `package-scripts.yml` file as well.

What about `nps init --type=yaml`?
Feel free to pick it up, otherwise I'll probably be able to do this some time this weekend. 😄

sxn:
**What**: #66 

**Why**: in order to allow those who prefer YAML to JS to use `nps` in their existing projects

**How**: by checking for the presence of the type modifier and, if present, leveraging [js-yaml](https://github.com/nodeca/js-yaml) to convert & dump the user's configuration

allow using the `package.json` of an existing project to generate a `package-scripts.yml` file in

order to make it easier to start using `nps`

closes #66

### DIFF
--- a/src/bin-utils/initialize/index.test.js
+++ b/src/bin-utils/initialize/index.test.js
@@ -3,7 +3,7 @@ import {resolve} from 'path'
 import {readFileSync} from 'fs'
 import {spy} from 'sinon'
 
-test('normal case initialize', () => {
+test('initialize JS normally', () => {
   const packageScriptsDestination = resolve('./src/bin-utils/initialize/fixtures/package-scripts.js')
   const packageJsonDestination = resolve('./src/bin-utils/initialize/fixtures/_package.json')
   const expectedPackageScripts = readFileSync(resolve('./src/bin-utils/initialize/fixtures/_package-scripts.js'), 'utf-8')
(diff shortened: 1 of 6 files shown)


src/bin-utils/initialize/index.test.js
```diff
@@ -21,14 +21,46 @@ test('normal case initialize', () => {
 
   initialize()
 
-  const [packageScriptsDestinationResult, packageScriptsStringResult] = mockWriteFileSync.firstCall.args
-  const [packageJsonDestinationReuslt, packageJsonStringResult] = mockWriteFileSync.secondCall.args
+  const [packageScriptsDestinationResult, packageScriptsStringResult] = mockWriteFileSync.secondCall.args
+  const [packageJsonDestinationResult, packageJsonStringResult] = mockWriteFileSync.firstCall.args
```
kentcdodds: lol, you fix so many of my typos. Thanks!

src/bin-utils/initialize/index.test.js
```diff
@@ -21,14 +21,46 @@ test('normal case initialize', () => {
...
+test('initialize YML normally', () => {
+  const packageScriptsDestination = resolve('./src/bin-utils/initialize/fixtures/package-scripts.yml')
+  const packageJsonDestination = resolve('./src/bin-utils/initialize/fixtures/_package.json')
+  const expectedPackageScripts = readFileSync(resolve('./src/bin-utils/initialize/fixtures/_package-scripts.yml'), 'utf-8')
+  const mockWriteFileSync = spy()
+  const mockFindUpSync = spy(file => {
+    if (file === 'package.json') {
+      return packageJsonDestination
+    }
+    throw new Error('Should not look for anything but package.json')
```
kentcdodds: oh, nice one 👍

src/bin/nps.js
```diff
@@ -102,6 +102,24 @@ function onInit() {
   ))
 }
 
+function getConfigType() {
+  let configType
+
+  if (includes(process.argv, '-t')) {
```
kentcdodds: If we're going to allow `-t` then we should probably document that. Personally, I'd be good to not allow `-t` and just accept `--type`

src/bin-utils/initialize/index.test.js
sxn: mostly for fear of not having introduced them by mistake :D

src/bin/nps.js
```diff
@@ -102,6 +102,20 @@ function onInit() {
   ))
 }
 
+function getConfigType() {
+  let configType
```
kentcdodds: I think this could be written a little more simply:

```javascript
return includes(process.argv, '--type') ? process.argv[indexOf(process.argv, '--type') + 1] : 'js'
```

Though... I'm not sure that's actually _simpler_... Just more terse. What do you think?

sxn: I was also not really satisfied with how it looked, so I think the ternary's the better way to go :)

src/bin/nps.js
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 /* eslint no-process-exit: "off" */
 import findUp from 'find-up'
-import {merge, includes} from 'lodash'
+import {merge, includes, indexOf, isUndefined} from 'lodash'
```
kentcdodds: whoops, `isUndefined` is no longer used 😅

sxn: oh god :D

kentcdodds:
LGTM

sxn:
Great! :shipit: 

Would you mind opening a ticket to discuss what could be done about moving to [yargs](https://www.npmjs.com/package/yargs) (or something else if necessary)? I'd be interested to see what your thoughts on that are so far.

PR #72, opened 2016-10-13
chore(yarn): add yarn support

kentcdodds:
Right now, I'm opening this against the upgrade-deps PR. Before we merge this, let's get that other one merged, then we'll merge this into master.

Also, I think something may be broken, so I'll have to come back to this if the build breaks.


.travis.yml
```diff
@@ -2,24 +2,26 @@ sudo: false
...
 node_js:
   - '6'
   - '4'
 before_install:
-  - npm i -g npm@^3.0.0
-before_script:
-  - npm prune
+  - time npm i yarn ---cache-min 99999999
+install:
+  - time $(npm bin)/yarn
```
tleunen: `$(npm bin)/yarn`? I don't see it in the deps?

tleunen:
Any reason why you didn't move to yarn? Is it not stable enough?

gunnx:
Yarn looks interesting, I'm hoping to try it out soon.

kentcdodds:
Yeah, I'm still planning on doing this eventually, but open source modules should not commit the `yarn.lock` file (to make sure they're always tested with the latest versions of modules which users will be downloading when they install your package). So I was just cleaning things up :)

PR #75, opened 2016-10-27
feat(help): Improve output of the help text to show nested scripts

### LINKED ISSUE #73: Indent child task descriptions when running help command
- `p-s` version: 3.0.1
- `node` version:  6.9.1
- `npm` version:  3.10.8

``` javascript
https://github.com/kentcdodds/p-s/blob/master/src/bin-utils/index.js
```

The command executed:

``` console
./node_modules/p-s/dist/bin/nps.js help
```

The output:

``` console
dev - DEVELOPMENT: Build CSS, build and run server, build, serve and watch client - npm start dev.build-sass,dev.runServerRunClientAndWatchSass
dev.buildServerAndStart - npm start dev.build,start

```

Problem description:
Tasks that are namespaced do not render out indented.

Suggested solution:
Output any nested tasks with spaces to indicate they are children of parent task

### LINKED ISSUE #74: Hidden scripts error when trying to run
- `p-s` version: 3.0.1
- `node` version:  6.9.1
- `npm` version:  3.10.8

``` javascript
https://github.com/kentcdodds/p-s/blob/master/src/resolve-script-object-to-string.js
```

Config extract:

``` javascript
dev: {
    build: {
      description: '',
      script: `NODE_ENV=development webpack --config ${webpackServerConfig}`,
      hiddenFromHelp: true
    },
...
```

The command executed:

``` console
nps dev.build
```

The output:

``` console
....
Scripts must resolve to strings. There is no script that can be resolved from "dev.build" https://github.com/kentcdodds/p-s/blob/v3.0.1/other/ERRORS_AND_WARNINGS.md#missing-script
....
```

Problem description:
Assume that the expected behaviour of flag `hiddenFromHelp` is to only remove it from the help list and not stop you running. 

Suggested solution:
Not sure I fully understand the code but looks like `resolvePlainObjectToScript` is returning `undefined` if the flag is set, though looks like this function is used by help and run tasks.

gunnx:
**What**:
The output for the help command has been updated to show nested format

**Why**:
To make it clear to users which scripts are children of other scripts

**How**:

Summary: Render out the script tasks as a tree structure, uses lib 'ascii-tree'

The existing function to `getAvailableScripts` was updated to include a parameter to indicate the indent level of a task, as the function recurses the level is incremented. 
Using 'ascii-tree' library we use a special character '#' to indicate indent level.

Closes #73


cli-test/__snapshots__/cli.test.js.snap
```diff
@@ -49,10 +49,9 @@ Object {
     -r, --require <module>                      Module to preload
 
 Available scripts (camel or kebab case accepted)
-
-test - echo \"test script\"
-lint - echo \"lint.default\"
-lint.sub.thing - this is a description - echo \"deeply nested thing\"
+├─ test - echo \"test script\"
+└─ lint - echo \"lint.default\"
+   │  └─ lint.sub.thing - this is a description - echo \"deeply nested thing\"
```
kentcdodds: What's this extra line for? It looks kind of odd to me. Can we make that not appear like that?

src/bin-utils/index.js
```diff
@@ -169,10 +173,12 @@ function getAvailableScripts(config, prefix = []) {
     const prefixed = [...prefix, key]
     if (scriptObj) {
       const {description, script} = scriptObj
-      scripts = [...scripts, {name: prefixed.join('.'), description, script}]
+      const name = prefixed.join('.')
+      const indent = indentCharacter.repeat(indentLevel)
```
kentcdodds: Hooray for ES6!

src/bin-utils/index.test.js
```diff
@@ -182,15 +182,14 @@ test('help: formats a nice message', () => {
...
-${colors.green('build.x.y')} - ${colors.white('build X-Y')} - ${colors.gray('echo "build x-y"')}
-${colors.green('foobar')} - ${colors.gray('echo "foobar"')}
+├─ ${colors.green('foo')} - ${colors.white('the foo script')} - ${colors.gray('echo "foo"')}
+├─ ${colors.green('bar')} - ${colors.white('stuff')} - ${colors.gray('echo "bar default"')}
+│  ├─ ${colors.green('bar.baz')} - ${colors.gray('echo "baz"')}
+│  └─ ${colors.green('bar.barBub')} - ${colors.gray('echo "barBub"')}
+├─ ${colors.green('build')} - ${colors.gray('webpack')}
+│  └─ ${colors.green('build.x')} - ${colors.white('webpack with x env')} - ${colors.gray('webpack --env.x')}
+│     └─ ${colors.green('build.x.y')} - ${colors.white('build X-Y')} - ${colors.gray('echo "build x-y"')}
+└─ ${colors.green('foobar')} - ${colors.gray('echo "foobar"')}
```
kentcdodds: This output looks pretty great.

src/bin-utils/index.js
```diff
@@ -4,13 +4,15 @@ import {remove, includes, isPlainObject, isEmpty} from 'lodash'
...
 import {safeLoad} from 'js-yaml'
+import tree from 'ascii-tree'
 
 import getLogger from '../get-logger'
 import {resolveScriptObjectToScript} from '../resolve-script-object-to-string'
 import initialize from './initialize'
 import {default as autocomplete, install as installAutocomplete} from './autocomplete'
 
 const log = getLogger()
+const indentCharacter = '#'
```
kentcdodds: I'm confused by this. I don't see the character `#` in any of the output. Is this something special to `ascii-tree` or something?

tleunen:
@kentcdodds I have quite a few scripts, with 2 or 3 subscripts. I can see the benefit of having them indented, to quickly see the "main" scripts and their subscripts. But at the same time, we are repeating the full command (for clarity) so I don't know if it give a huge benefit after all.

Would it be possible to have an example in a terminal? I can fetch the PR and test against one of my project too if you don't have any test project @gunnx

gunnx:
@tleunen Yes please, my project doesn't have much more then the p-s project itself.

cli-test/__snapshots__/cli.test.js.snap
gunnx: I'm not sure, but I guess that's the behaviour of ascii tree.

src/bin-utils/index.js
gunnx: Yes a special character is used to indicate the indent level, it doesn't have to be # any character could be used.

kentcdodds: Gotcha. Could we use just `' '`?

boneskull:
I am ambivalent here. It may be easier to see how commands relate at the cost of potentially soft-wrapping.

gunnx:
@kentcdodds I should have said as well this is my first PR as I know you are involved in support that.

kentcdodds:
Let me try this out on my work project that has a lot of scripts and see how it looks.

kentcdodds:
Alrighty! I've published a beta version of your changes and people can try it out by installing `p-s@beta`. Looks pretty good, but I think I've found a bug:

screen shot 2016-10-29 at 9 30 01 pm

You'll see that the `watch` scripts almost look like they're coming out of the `flow` script. But in reality there's just not a `watch` script or a `watch.test` script for that matter. So we'd want to de-indent those I think.

gunnx:
@kentcdodds Are you able to post the package-scripts file too? Just wondering why you'd have sub-sub tasks without any parent tasks?

tleunen:
I'm experiencing the same thing that @kentcdodds reported.

screen shot 2016-11-01 at 4 25 52 pm

It seems it happens when tasks don't have any `script` @gunnx

gunnx:
@tleunen Hi what does that look like in the package-scripts file?

tleunen:
Like this:
screen shot 2016-11-01 at 4 41 52 pm

The result of the `getCopyImgCmd` is the command executed.

gunnx:
@tleunen Yes thanks, I can see that resolveScriptObjectToString will return undefined if it can't find a script. Is this expected behaviour? Should script be made to be required if you define it as an object

kentcdodds:
> Should script be made to be required if you define it as an object

No, I think that @tleunen's usage is totally valid and we should probably make that look good in this output.

gunnx:
@kentcdodds Ok I think I have a way, so are you saying if there is a nested script but it has no parent script then it should not be indented.

kentcdodds:
Hey @gunnx, I've been using this for a little while, and I'm not convinced that it improves the readability of the output... Especially when wrapping the text. I don't think that I'm going to merge this one. I'm really sorry about that. If you want to give your hand in another PR on the repo, I'll definitely accept a PR that fixes this but you reported: https://github.com/kentcdodds/p-s/issues/74

gunnx:
Ok sure, I'll take a look at the other issue

kentcdodds:
Again, I'm sorry your first pull request didn't work out. But don't give up! This was a great pull request! You can do great things in open source 😀

gunnx:
No problem, it was good to go through the process :)

PR #77, opened 2016-11-09
chore(devDeps): add missing peer dep of babel-plugin-module-resolver

boneskull:
`eslint-import-resolver-babel-module` was complaining

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "babel-cli": "6.11.4",
     "babel-eslint": "6.1.2",
     "babel-jest": "15.0.0",
+    "babel-plugin-module-resolver": "2.3.0",
     "babel-preset-es2015": "6.9.0",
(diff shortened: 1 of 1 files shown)


kentcdodds:
Yeah, that happens sometimes. It's kinda weird. Thanks! :shipit:

tleunen:
Author of babel-plugin-module-resolver and eslint-import-resolver-babel-module here, sorry for this. I made a mistake with the recent update and I should have released a major because the signature of an exported function changed :/ But because I made an update on both packages at the same time I thought it was ok but definitely not because I haven't thought of the case where a dependency of your project could also use the babel plugin.
Definitely lesson learned here.

Sorry again

kentcdodds:
Thanks for the tip @tleunen! And thanks for the `module-resolver` plugin. One of my favorites!

PR #83, opened 2016-12-30
Add life cyle pre and post commands

rrag:
There are breaking tests, I am not sure how to fix 2 of them, feedback welcome

**What**:
Implement lifecycle pre and post commands for scripts, based on
https://github.com/kentcdodds/p-s/issues/67#issuecomment-251704554

**Why**:
Add a new feature

**How**:
By returning a `script` of shape

```
{
  pre: ...,
  script: ...,
  post: ...,
}
```
object instead of a string, and running the commands if present in the order of `pre` followed by `script` and then `post`

### DIFF
--- a/src/get-script-to-run.js
+++ b/src/get-script-to-run.js
@@ -1,6 +1,6 @@
 import {each, cloneDeep, isPlainObject} from 'lodash'
 import prefixMatches from 'prefix-matches'
-import resolveScriptObjectToString from './resolve-script-object-to-string'
+import resolveScriptObjectToLifecycleObject from './resolve-script-object-to-string'
 import kebabAndCamelCasify from './kebab-and-camel-casify'
 
 export default getScriptToRun
@@ -9,18 +9,18 @@ function getScriptToRun(config, input) {
   config = kebabAndCamelCasify(config)
   // remove the default objects/strings so we can check if the prefix works with another script first
   const defaultlessConfig = removeDefaults(cloneDeep(config))
-  const scriptString = getScriptString(defaultlessConfig, input)
-  if (scriptString) {
+  const scriptString = getScriptStringWithLifecycle(defaultlessConfig, input)
+  if (scriptString && scriptString.script) {
     return scriptString
   } else {
     // fallback to the defaults if no other script was found with the given input
-    return getScriptString(config, input)
+    return getScriptStringWithLifecycle(config, input)
   }
 }
 
-function getScriptString(config, input) {
+function getScriptStringWithLifecycle(config, input) {
   const [script] = prefixMatches(input, config)
-  return resolveScriptObjectToString(script)
+  return resolveScriptObjectToLifecycleObject(script)
 }
 
 function removeDefaults(object) {
--- a/src/get-script-to-run.test.js
+++ b/src/get-script-to-run.test.js
@@ -1,30 +1,30 @@
 import getScriptToRun from './get-script-to-run'
 
 test('allows a prefix to be provided', () => {
-  const script = getScriptToRun({build: 'stuff'}, 'b')
+  const {script} = getScriptToRun({build: 'stuff'}, 'b')
   expect(script).toBe('stuff')
 })
 
 test('allows a multi-level prefix to be provided', () => {
-  const script = getScriptToRun({build: {watch: 'watch stuff'}}, 'b.w')
+  const {script} = getScriptToRun({build: {watch: 'watch stuff'}}, 'b.w')
   expect(script).toBe('watch stuff')
 })
 
 test('falls back to using `get` for the full name if no prefix is provided', () => {
-  const script = getScriptToRun({build: {watch: 'watch stuff'}}, 'build.watch')
+  const {script} = getScriptToRun({build: {watch: 'watch stuff'}}, 'build.watch')
   expect(script).toBe('watch stuff')
 })
 
 test('can accept snake-case representation of a camelCase name', () => {
-  const script = getScriptToRun({checkCoverage: 'checking coverage'}, 'check-coverage')
+  const {script} = getScriptToRun({checkCoverage: 'checking coverage'}, 'check-coverage')
   expect(script).toBe('checking coverage')
 })
 
 test('fallsback to `default` if no prefix is found', () => {
   const scripts = {foo: {default: 'echo "default"', dee: 'echo "dee"'}}
-  const usesDefault = getScriptToRun(scripts, 'foo')
-  const defaultIsPrefixFallback = getScriptToRun(scripts, 'foo.def')
-  const script = getScriptToRun(scripts, 'foo.de')
+  const {script: usesDefault} = getScriptToRun(scripts, 'foo')
+  const {script: defaultIsPrefixFallback} = getScriptToRun(scripts, 'foo.def')
+  const {script} = getScriptToRun(scripts, 'foo.de')
 
   expect(usesDefault).toBe('echo "default"')
   expect(defaultIsPrefixFallback).toBe('echo "default"')
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,7 @@
 import spawn from 'spawn-command-with-kill'
 import Promise from 'bluebird'
 import colors from 'colors/safe'
-import {isString, clone} from 'lodash'
+import {isString, isUndefined, clone} from 'lodash'
 import {sync as findUpSync} from 'find-up'
 import managePath from 'manage-path'
 import arrify from 'arrify'
@@ -12,6 +12,17 @@ import getLogger from './get-logger'
 const NON_ERROR = 0
 
 export default runPackageScripts
+
+/*
+  scriptConfig: Object which is exported by package-scripts.js
+  scripts: the scripts (yes array) which are to be executed now,
+    eg. for `nps foo` scripts is ["foo"]
+    eg. for `nps foo,bar,baz` scripts is ["foo", "bar", "baz"]
+  args: args passed to the run command
+    eg. for `npm start foo -- --arg1 --arg2` and if scriptConfig is { foo: "echo bar" }
+    the args will be ["arg1", "arg2"]
+  options: { parallel: true/false, logLevel: one of "error", "warn", "info", silent: true/false }
+*/
 
 function runPackageScripts({scriptConfig, scripts, args, options = {}}) {
   if (scripts.length === 0) {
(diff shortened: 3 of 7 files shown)


kentcdodds:
This is looking great! I think some things could be cleaned up with async/await. Mind giving that a try? I hope you don't mind iterating on this a bit!

kentcdodds:
Looks like there may be some issues. See the relevant CI build. Also, the test with all the commands is slightly confusing. If you could split things up a bit that might help.

kentcdodds:
Thank you a bunch for working on this!

rrag:
> I think some things could be cleaned up with async/await

I have not used `async await` would love to learn and contribute here. Are you suggesting something like this?

```js
const preResonseCode = runPackageScript({script: pre, options, scriptName: preScriptName, args})
const preResonseCode = runPackageScript({script, options, scriptName, args})
const preResonseCode = runPackageScript({script: post, options, scriptName: postScriptName, args})
```

and in `runPackageScript` just resolve immediately if `script` is undefined

```js
async function runPackageScript(...) {
  ...
  const promise = await new Promise(...) // resolve the promise immeidiately if `script` is undefined
  return promise
}

```

Unrelated:
I had some initial trouble understanding the project so I wanted to use the Jest feature `test.only` on a few tests to understanding how the tests work, but it seems the version of jest used here does not support `.only` however when I upgraded, none of the tests ran, I could not figure out why, need to dig deeper.

kentcdodds:
Thanks for digging into that. Yeah, basically what you showed is how it'll look (except with more `await`, and perhaps a `try/catch` or wrap it all in a function and a `.catch` on a call to that function). In any case. I've got [this PR](https://github.com/kentcdodds/p-s/pull/84) which should upgrade everything. Hopefully someone can review/merge it soon!

kentcdodds:
Alrighty, if you could [rebase](https://egghead.io/lessons/javascript-how-to-rebase-a-git-pull-request-branch) this with what's on master then you should have the latest and greatest of everything! Thank you :)

rrag:
Thank you for the help, sorry for the delay I will look at this early next week.

rrag:
sorry for delay.

I tried with async await, by making
```
async function runPackageScriptWithLifecycle(...) {
  await runPackageScript(pre...)
  await runPackageScript(script...)
  await runPackageScript(post...)
}
```

but doing that means that `runPackageScriptWithLifecycle` is not returning a promise. the parallel option depends on [each script being a promise](https://github.com/kentcdodds/p-s/blob/master/src/index.js#L39-L41). Also that expects an `abort` method on the promise to be called if any of the scripts running in parallel fails. So to be able to provide the abort feature I had to make the `runPackageScriptWithLifecycle` return a promise.

Also I commented [this](https://github.com/rrag/p-s/blob/feat/lifecycle/src/index.test.js#L197-L198)

if I make that test run as `.only` then it passes, but it fails when the only is removed, so I had to comment that assertion to make it work in all cases. Appreciate your help in looking at this one issue before merging

kentcdodds:
Hey, thanks for your effort on this. I'm having second thoughts about it though... Can you give me a scenario where this would really be better than just using JavaScript to create these relationships in `package-scripts.js` rather than having the implicit relationship as part of the tool? I'm just starting to question whether it's worth the complexity...

gunnx:
I like the idea of this, not sure about the complexity it adds to p-s though. It's built in feature of NPM scripts, so I can see people wanting to use the same style.

kentcdodds:
Yeah, the main use of npm script hooks is to hook into when npm does something interesting (install/publish), which is not something we could really do. The other use is for custom script lifecycles, but that's only necessary because the static nature of JSON. I think that people can easily get around that limitation without much confusion. So I vote let's not push this feature... Anyone else?

rrag:
The reason I started to work on this is because one of my `package-scripts.js` is becoming increasingly complex

https://gist.github.com/rrag/94847b9559085dbc9568ce078f5c0af5 (I might delete this gist later)

I had created, to deal with build and deploy of different parts of an app. there are different parts of the app which require varying levels of build tooling.

and I really wanted to do something like

```js
deploy: {
  pre: "nps build",
  script: "nps aws.updateFunction"
  post: "nps aws.postDeploy"
},
postDeploy: {
  pre: ""nps aws.testLambdaUsingInvokeFunction", // test the function in aws lambda
  script: "nps promoteFunction" // creates a new version and changes the tag so api gateway uses this new version
}
```
this way I could have individual smaller scripts, and then these bigger orchestrator scripts which do multiple things using lifecycle

`package.json` is no way to manage such complex scripts, where there are variable names, slightly different commands between windows and others (and more importantly lack of comments). This is a great package, I don't think this package gets the attention that it deserves.

I don't think this pr increases the complexity too much. just `+160 −31`, so given the way it could simplify a complex build processes, and the similarity between npm scripts make for a compelling case to merge this in.

rrag:
If I may, parallel option is the reason for increased complexity. The tests for parallel option are not easy to understand the use of spy, it took me a while to understand what was going on. the use of `abort` method on a promise was not aparent and took me a while to figure out. Removing the parallel feature and adding lifecycle would be my suggestion. If someone needs parallel they could use something like `concurrently`. Although I am certainly biased as I never use the parallel option :D

gunnx:
I had issues with the parallel option, so instead use `parallelshell` so
have no issue with removing that.

kentcdodds:
> I don't think this package gets the attention that it deserves.

Thanks! I agree!

> I don't think this pr increases the complexity too much. just `+160 −31`

I'm less worried about complexity in the tool implementation itself, and more concerned about the complexity of learning and understanding the tool. I'm worried about having too many concepts. I looked over your scripts and I don't really see how this could reduce complexity in your scripts. In some of my `package-scripts.js` files, I have functions that I use to cut the cruft in several scripts. I think you may benefit from this.

As for the complexity of `parallel` and the ability to just use `concurrently`, that's intriguing. I might consider removing `--parallel` in favor of encouraging the use of `concurrently`. If someone wants to make a PR for the documentation changes (I want to see an example of what the scripts might look like) then we can have a conversation about that. Reducing the API surface area is a good thing :)

rrag:
I respect your opinion, and if you decide to not merge, that is alright, but I want to register my 2 cents.

Features get added in any tool over time, I am not sure it should be considered as increased surface area. if there are multiple ways of doing the same thing, like `React.createClass` and `extends React.Component` it causes confusion and is a reason to remove one, but this is a new feature. I am not convinced your argument holds merit in this case.

kentcdodds:
I'm going to put this one up to a vote. For clarity, here's what the feature is: Lifecycle `pre` and `post` scripts similar to how npm lifecycle scripts work.

Arguments for:
- It could simplify scripts (so you have fewer `&&` in your scripts)

Arguments against:

- It wont actually work for the npm lifecycle (like a script called `prepublish` wouldn't happen before `npm publish` actually takes place). So it could cause confusion in that way (perhaps we could log a warning if you try to use lifecycle script names which you could suppress in `options`)
- It would add more API surface area. More to learn for the users of this tool. And more complexity in the tool itself. It would be reasonable to use JavaScript functions in `package-scripts.js` to simulate this kind of API yourself (perhaps we could expose such functions via a node API).

---

Please use GitHub emoji reactions on this comment to vote.

- :+1: means "Yes, let's add this feature"
- :-1: means "No, let's not add this feature"
- :confused: means "I'm unsure"
- :smile: means "I don't really care either way"

I'll check back in ~24 hours. I'll try to reach out to as many people as use the tool as possible. Here's a start:

cc/ @DavidWells, @abhishekisnot, @rowanoulton, @giladgo, @tim-mcgee, @nkbt, @tleunen, @Hypercubed, @jisaacks, @boneskull, @RobinMalfait, @edm00se, @SamVerschueren, @sxn, @gunnx, @martellaj, @msegado, @beeman, @elijahmanor

Feel free to comment if you have additional input/thoughts.

tleunen:
Not sure it's a good idea. It will add a bit of complexity in the project for not a huge gain in the end.
pre/post scripts could be done easily by writing your script like that:
```
script: [
  'doPre',
  'currentScript',
  'doPost',
].join(' && '),
```

elijahmanor:
I voted 😕 mainly because it does add complexity for something I'm not sure how many users will want the feature.

Since the *idea* is supported by native `npm scripts` I can see not having the feature being a means of confusion if someone converts their existing `package.json` to `p-s` only to find their pre/post scripts no longer work. Of course you can rearrange the scripts to work, but it could be a hinderance. 

Unless I've overlooked it, I don't see any reference in the docs that `p-s` does *not* support pre/post commands. That information might be nice to add somewhere in the documentation (`README.md`) if this feature isn't added since it may be expected by the user.

kentcdodds:
Based on the voting, I'm going to close this PR. But I would appreciate another PR that adds some documentation as @elijahmanor mentions. So when people look for lifecycle hooks, they can see an example of how to accomplish it (like @tleunen mentioned). Any takers?

kentcdodds:
And @rrag, thank you so much for your hard work on this. I really do appreciate it. I'm going to add you to the contributors table for all that work you put in. Thank you very much!

benmonro:
seems to me that at the very least init script should import scripts from package.json so that they work the same under p-s.

i.e. if I run `nps init` on the following package.json

```json
scripts:{
  "prething": "echo pre-thing",
   "thing":"echo thing"
}
```

then I should get the following output (in yaml)
```yaml
scripts:
    prething: echo pre-thing
    thing: nps prething && echo thing
```

kentcdodds:
Agreed. Could you open a new issue for us to track that?

benmonro:
https://github.com/kentcdodds/p-s/issues/108 @kentcdodds

Zerogiven:
Hmm i'm not sure if it confuses users that if you have for example something like "prepublish" inside nps scripts and execute "npm publish" it is normal that it does not execute prepublish from there...

"npm start" also does not execute "nps start" thats why we have the "start" script at package.json. It's the same like "npm test" and a few other keywords.

So if for example i want execute prepublish like in your example user could do following:

package.json
```json
{
  "scripts": {
    "publish": "nps publish"
  }
}
```

at package-scripts.js
```js
module.exports = {
  publish: {
    pre: ".....",
    default: ".....",
  }
}
```

So i can make "npm publish" with pre task execution.

The only thing what i'm not 100% sure about is which execution order you have when you add prepublish to package.json too. It should be like.

```bash
npm publish
```

which executes:

```bash
npm run prepublish && npm run publish
```

npm run publish executes:

```bash
nps run publish.pre && nps run publish
```

I also don't think that it makes nps more complicated cause all tasks in package-scripts with post or pre attribute are only executed at nps package-scripts. package.json post or pre scripts are only executed at npm run scripts.

kentcdodds:
Thanks for your input @Zerogiven. But I'm not convinced that justifies the feature. The less magic behavior in `nps` the better IMO.

boneskull:
agree. KISS

PR #87, opened 2017-01-10
feat(cli): ask for confirmation before overwriting an existing config file

### LINKED ISSUE #86: nps init should ask if you want to override an existing package-scripts.js
First of all thanks for creating this great utility. It makes my work a lot easier :)

- `p-s` version: 3.0.3
- `node` version: v7.2.0
- `npm` version: 3.10.10

```javascript
# Any customized script
```

The command executed:

```console
$ nps init
```

The output:

```console
Your scripts have been saved at /private/tmp/p-s-is-awesome/package-scripts.js
Check out your scripts in there. Go ahead and update them and add descriptions to the ones that need it
Your package.json scripts have also been updated. Run `npm start help` for help
You may also want to install the package globally and installing autocomplete script. You can do so by running
  npm install --global p-s
  nps completion <optionally-your-bash-profile-file>
The bash profile file defaults to ~/.bash_profile for bash and ~/.zshrc for zsh
```

Problem description:

When there is an existing `package-scripts.js` in the project the command `nps init` will override it with the defaults. 

I'm not sure if this is by design but if so I think it makes sense to have a confirmation that asks the user for a confirmation to override it. 

Suggested solution:

I think there are generally two solutions to consider:
- Ask for confirmation to override the existing file.
- Skip creating the file when it exists, and inform the user to remove the existing file.

beeman:
**What**:
Fixes #86 

**Why**:
See #86 

**How**:

I've looked for ways to get the confirmation done with node built-in packages (like `readline`) or existing dependencies in this project but could not find any. That's why I opted to depend on [readline-sync](https://www.npmjs.com/package/readline-sync) which I think is a clean way of doing it. 

Let me know if you have other suggestions to do this.  

Not sure if/how I should test this, as `src/bin/nps.js` currently does not have any tests.


src/bin/nps.js
```diff
@@ -87,6 +88,11 @@ function getPSConfigFilepath() {
 }
 
 function onInit() {
+  if (getPSConfigFilepath()) {
+    if (!keyInYN(colors.yellow(`Do you want to overwrite your existing config file?`))) {
```
kentcdodds: Should we add `(Y/N)`? So people know what to type?

src/bin/nps.js
```diff
@@ -87,6 +88,11 @@ function getPSConfigFilepath() {
...
+      process.exit(1)
```
kentcdodds: Could we use `FAIL_CODE` here instead of hard-coding `1`?

src/bin/nps.js
beeman: The library adds that for us:
screen shot 2017-01-10 at 00 41 48

kentcdodds: Excellent!

src/bin/nps.js
beeman: Good catch, fixed.

PR #91, opened 2017-02-05
docs(README): update video screenshot to egghead.io video

kentcdodds:
**What**: docs(README): update video screenshot to egghead.io video

**Why**: Because @elijahmanor's video is awesome

**How**: Screenshot a good spot in the video, shrink it down to 2x 700px (which is the max width of the image in the README, for retina). Then use tinypng.com to make it as small as possible.

I also changed the short URL (the one we're using for the link) to go to
@elijahmanor's awesome Egghead.io lesson about it!)


kentcdodds:
Anyone have ideas for why the build is busted?

tleunen:
I have restarted the job because the error was really weird... And everything's fine now...

kentcdodds:
Thanks!

Issue #70, opened 2016-10-04
Move to yargs


kentcdodds:
Commander is cool and all, but there's some hacky things we have to do to get our CLI to do what we want. It'd be sweet to use [yargs](https://github.com/yargs/yargs) which is much more flexible.

All changes should happen in [here](https://github.com/kentcdodds/p-s/blob/master/src/bin/nps.js). Should hopefully be able to just use it similar to how I've used it in [split-guide](https://github.com/kentcdodds/split-guide) [here](https://github.com/kentcdodds/split-guide/blob/fb4b2a2ebc1fb8c3c010c2af1318861b8bb1bb13/src/bin/index.js#L6-L30) :+1:

sxn:
That's a pretty solid starting point! 👍 
Some things that should change [here](https://github.com/kentcdodds/p-s/blob/master/src/bin-utils/index.js#L89) as well, from the looks of it, right?

kentcdodds:
Yes, probably so...

nkbt:
I actually found `nomnom` to be the best. Just in case you've never heard of it http://npm.im/nomnom

kentcdodds:
I've never used `nomnom`, but I have used `yargs` and liked it pretty well. It's also much more widely used. I think I'd like to stick with that one personally. What do you think @sxn?

sxn:
Hm. Actually I hadn't played around with either of those, so I can't form an informed opinion. I'll briefly look into both before diving in and come back. Without looking into them I'd probably choose {{yargs}} just because there would be at least one person who knows this codebase *and* the library we'd be introducing. :D

PR #94, opened 2017-02-06
remove parallel

kentcdodds:
**What**: Removes `parallel`

**Why**: Reduces complexity of the tool and its implementation.

**How**: It's still a work in progres...

This will be a breaking change... Still evaluating.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ module.exports = {
     default: 'node index.js',
     lint: 'eslint .',
     test: {
-      // learn more about Jest here: https://kcd.im/egghead-jest
+      // learn more about Jest here: https://facebook.github.io/jest
       default: 'jest',
       watch: {
         script: 'jest --watch',
@@ -52,10 +52,12 @@ module.exports = {
       }
     },
     build: {
+      // learn more about Webpack here: https://webpack.js.org/
       default: 'webpack',
       prod: 'webpack -p',
     },
-    validate: 'nps --parallel lint,test,build',
+    // learn more about concurrently here: https://npm.im/concurrently
+    validate: 'concurrently "nps lint" "nps test" "nps build"',
   },
 }
 ```
@@ -75,7 +77,7 @@ scripts:
     build:
         default: webpack
         prod: webpack -p
-    validate: nps --parallel lint,test,build
+    validate: concurrently "nps lint" "nps test" "nps build"
 ```
 
 To use `p-s`, it's recommended that you either install it globally (`npm i -g p-s`) or add `./node_modules/bin` to your
@@ -84,7 +86,7 @@ To use `p-s`, it's recommended that you either install it globally (`npm i -g p-
 Then you can run:
 
 ```console
-p-s --help
+nps --help
 ```
 
 Which will output:
@@ -97,7 +99,6 @@ Which will output:
     -h, --help                                  output usage information
     -V, --version                               output the version number
     -s, --silent                                Silent nps output
-    -p, --parallel <script-name1,script-name2>  Scripts to run in parallel (comma seprated)
     -c, --config <filepath>                     Config file to use (defaults to nearest package-scripts.yml or package-scripts.js)
     -l, --log-level <level>                     The log level to use (error, warn, info [default])
     -r, --require <module>                      Module to preload
@@ -109,7 +110,7 @@ test - jest
 test.watch - run in the amazingly intelligent Jest watch mode - jest --watch
 build - webpack
 build.prod - webpack -p
-validate - nps --parallel lint,test,build
+validate - concurrently "nps lint" "nps test" "nps build"
 ```
 
 **Because `p-s` is harder to type, it is recommended that you use the alias `nps` to interact with `p-s`, which is much
@@ -243,14 +244,6 @@ config).
 By default, `nps` will log out to the console before running the command. You can add `-s` to your command to silence
 this.
 
-##### -p, --parallel
-
-Run the given scripts in parallel. This enables handy workflows like this:
-
-```console
-nps -p lint,build,cover && nps check-coverage && nps report-coverage
-```
-
 ##### -c, --config
 
 Use a different config
@@ -283,7 +276,7 @@ nps lint --fix # --fix will be passed on to the lint script
 
 ##### scripts
 
-If you don't use `-p` (because you don't need parallelism) then you can simply provide the name of the script like so:
+To run a script, you simply provide the name of the script like so:
 
 ```console
 nps cover
@@ -377,6 +370,10 @@ Log levels available:
 
 ## FAQ
 
+### How do I do ___ ?
+
+Have you looked at the examples in [other/EXAMPLES.md][examples]?
+
 ### Why `npm start`?
 
 _Just to be clear:_ You do **not** have to use the `start` script. You can use whatever you like. But I recommend using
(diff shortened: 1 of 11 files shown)


tleunen:
😢 

Why removing it? Would be great to keep it but use `concurrently` under the hood instead?

kentcdodds:
I'm looking the reduced complexity from both an API standpoint as well as an implementation standpoint. It's altogether simpler and more focused. And we get even more features and functionality by deferring to another package for this. Still considering it.

kentcdodds:
I'd like to put this to a vote. For clarity, here's what's happening here:

This removes the `--parallel` feature from `p-s` and instead recommends usage of [`concurrently`](https://github.com/kimmobrunfeldt/concurrently) to have scripts that run at the same time.

Arguments for:

- Reduces the tool's API surface area. Less for people to learn about the tool.
- Simplifies the implementation (considerably) and pushes things off to another tool which can have more uses and is well tested.
- Follows the [DOTADIW](https://en.wikipedia.org/wiki/Unix_philosophy#Do_One_Thing_and_Do_It_Well) philosophy. Allows composability of existing tools that each "do one thing and do it well" which can be really nice and flexible.

Arguments against:

- Users have to install another dependency to get the parallel functionality.

---

Please use GitHub emoji reactions on this comment to vote.

- :+1: means "Yes, let's add this feature"
- :-1: means "No, let's not add this feature"
- :confused: means "I'm unsure"
- :smile: means "I don't really care either way"

I'll check back in ~24 hours. I'll try to reach out to as many people as use the tool as possible. Here's a start:

cc/ @DavidWells, @abhishekisnot, @rowanoulton, @giladgo, @tim-mcgee, @nkbt, @tleunen, @Hypercubed, @jisaacks, @boneskull, @RobinMalfait, @edm00se, @SamVerschueren, @sxn, @gunnx, @martellaj, @msegado, @beeman, @elijahmanor

Feel free to comment if you have additional input/thoughts.

src/index.js
```diff
@@ -18,11 +18,7 @@ function runPackageScripts({scriptConfig, scripts, args, options = {}}) {
     scripts = ['default']
   }
   const scriptNames = arrify(scripts)
-  if (options.parallel) {
-    return runParallel()
-  } else {
-    return runSeries()
-  }
+  return runSeries()
```
sxn: if we do end up removing it we should probably open a refactoring issue to clean this type of things up :)

beeman:
cc/ @mrfelton

README.md
```diff
@@ -75,7 +77,7 @@ scripts:
     build:
         default: webpack
         prod: webpack -p
-    validate: nps --parallel lint,test,build
+    validate: concurrently "nps lint" "nps test" "nps build"
```
sxn: Concurrently might prove [an extra headache for Cygwin users](https://github.com/kimmobrunfeldt/concurrently/issues/69#issuecomment-257698185), I guess. I'll try to see if there's any way to fix that though instead of using that as a basis for the decision at hand.

sxn:
I'd go with a 😕 for now. Reducing the surface area's great (as well as having less code to maintain :D) but I'm wondering if in the long run it won't complicate things for users more than it simplifies them.

README.md
```diff
@@ -44,18 +44,20 @@ module.exports = {
...
       }
     },
     build: {
+      // learn more about Webpack here: https://webpack.js.org/
       default: 'webpack',
       prod: 'webpack -p',
     },
-    validate: 'nps --parallel lint,test,build',
+    // learn more about concurrently here: https://npm.im/concurrently
+    validate: 'concurrently "nps lint" "nps test" "nps build"',
```
tleunen: This is quite verbose.
If this is not possible yet, could we support this syntax?
```
concurrently "nps lint test build"
```

beeman: @tleunen the issue is that we now have to rely on how concurrently works, and it expects complete commands, not just the parameters like you would with `nps -p`

kentcdodds: I don't think that we could support that syntax unfortunately. Because concurrently will spawn a process with the command `nps lint test build`. If you look at the additions in `EXAMPLES.md` you'll see that a function can be used to simplify things further. Perhaps we could use that example in the README instead?

gunnx:
Any reason for concurrently over npm-run-all?

kentcdodds:
> Any reason for concurrently over npm-run-all?

Unless I'm mistaken, `npm-run-all` will only run npm scripts. Because we don't actually put our scripts in `package.json` it's incompatibile.

When I use raw npm scripts, `npm-run-all` is one of my favorite packages :)

kentcdodds:
Sorry everyone, I had a copy/paste issue. I'm pretty sure the intent is still the same, but I updated the emoji voting option descriptions to make sure things were more clear.

kentcdodds:
Just updated this PR. Here's a little backstory:

- concurrency has this cool feature where it will allow you to label the output from each script (because normally the output is jumbled together and you have to kinda guess where the output is coming from).
- However, enabling this functionality caused the colors from the output to be removed :-( This made the output even harder to read than before.
- concurrency allows you to pass `--raw` and we're basically back to where we were with the `--parallel` flag where the output of the scripts is jumbled together. Status quo.
- [I dug into it a bit](https://github.com/kimmobrunfeldt/concurrently/issues/86) and found out that [`colors`](http://npm.im/colors) (the module we were using to color our output) wont color the output if it detects that you're in an environment that doesn't support color.
- I dug more and found out that [`chalk`](http://npm.im/chalk) (the de-facto standard for coloring output for the terminal these days) also attempts to detect whether to actually color the strings you give it. However, it also exposes some ways to override the behavior. Specifically, adding `FORCE_COLOR=true` will force chalk to color the text.
- So I'm going to see if we can add this capability to concurrency.
- Until that time, if you simply add a `process.env.FORCE_COLOR = true` to your `package-scripts.js` it will force chalk colors to show.
- However, we were using `colors` to color the output from `nps` (like `nps executing: script string`). So this PR replaces `colors` with the more popular (and widely supported) `chalk` module. Was pretty much just a find/replace operation.

Fun stuff!

README.md
kentcdodds: You can also check out the changes to `package-scripts.js` in this PR. I think it's pretty clean.

README.md
kentcdodds: Thanks for finding that. Seems like it wouldn't be too terrible to fix.

kentcdodds:
I should probably also mention that all the trouble I went through in the previous comment was so we could have even _more_ functionality with `concurrently` than we currently have now with `--parallel`. We don't have to do anything special to get parody functionality with `concurrently`.

kentcdodds:
Ok, based on the voting, I think we're going to make this happen. Regardless of whether you're in favor, I appreciate any input on how we can make this change as successful as possible. What can I change about this PR to improve things? Here are a few ideas:

1. Add the `concurrent` function (from `EXAMPLES.md` to the `README.md`) to inspire people to use JavaScript to make their scripts file cleaner and demonstrate that using concurrently doesn't have to be verbose.
2. Add a few helper functions people can require into their `package-scripts.js` to make doing this easier? Maybe? I'm not 100% sure I'm in favor of this.

Any other ideas?

Also, because this is a breaking change, I've set the base branch to `next` which is where I'll put any breaking changes we want to have go out at the same time. If anyone can think of any other improvements that would result in a breaking change, let me know! I want to avoid doing these major version bumps where possible. The plan is to merge things into `next` and when we're good and ready, we'll make a PR from `next` to `master`, merge it, and it'll get automatically released ✨

kentcdodds:
Just had an idea. If we see the `--parallel` flag, we should log a warning indicating that this was changed and linking to [`ERRORS_AND_WARNINGS.md`](https://github.com/kentcdodds/p-s/blob/master/other/ERRORS_AND_WARNINGS.md) which will describe how to get parallel capabilities.

kentcdodds:
With all the changes I'm making for #70, I'm going to do the `--parallel` warning later. I'll merge this into `next` now and keep working on #70 and #100.

PR #98, opened 2017-02-08
fix(load): allow function to be exported (as docs say)

### LINKED ISSUE #88: /bin-utils loadJSConfig() doesn't support package-scripts.js exporting a function
[`src/get-scripts-from-config.js`](https://github.com/kentcdodds/p-s/blob/master/src/get-scripts-from-config.js#L5) allows `package-scripts.js` to export a function:
```js
import {isPlainObject, isFunction} from 'lodash'

export default getScriptsFromConfig

function getScriptsFromConfig(scripts, input) {
  if (isPlainObject(scripts)) {
    return scripts
  } else if (isFunction(scripts)) {
    return scripts(input)
  }
  return {}
}
```

but [`src/bin-utils/index.js loadJSConfig()`](https://github.com/kentcdodds/p-s/blob/master/src/bin-utils/index.js#L28) does not:
```js
const loadJSConfig = getAttemptModuleRequireFn(function onFail(configPath, requirePath) {
  log.error({
    message: colors.red(`Unable to find JS config at "${configPath}". Attempted to require as "${requirePath}"`),
    ref: 'unable-to-find-config',
  })
  return undefined
})

...

function loadConfig(configPath) {
  if (configPath.endsWith('.yml')) {
    return loadYAMLConfig(configPath)
  }

  return loadJSConfig(configPath)
}
```

So when attempting to run a `default` script via command line, you get the generic help output with a "no scripts" warning:
```js
// package-scripts.js
exports.scripts = function (input) { ... }
```
```bash
$ nps

  Usage: nps [options]

  Options:

    -h, --help                                  output usage information
    -V, --version                               output the version number
    -s, --silent                                Silent nps output
    -p, --parallel <script-name1,script-name2>  Scripts to run in parallel (comma seprated)
    -c, --config <filepath>                     Config file to use (defaults to nearest package-scripts.yml or package-scripts.js)
    -l, --log-level <level>                     The log level to use (error, warn, info [default])
    -r, --require <module>                      Module to preload

There are no scripts available
```

kentcdodds:
**What**: closes #88

**Why**: Because we document that you can export a function that returns
a configuration, but it doesn't actually work.

**How**: Add some logic to `loadConfig` in `src/bin-utils/index.js` to
handle the case when the config is a function. Also, we need to pass the
input along, so I had to make a few other changes as well. Also added
fixtures and tests for this.

This also adds @erikfox to the contributors table for reporting the bug! Thanks!


kentcdodds:
Anyone wanna review this really quick?

PR #99, opened 2017-02-08
fix(init): don't fail when there are no scripts

### LINKED ISSUE #89: Error thrown if "scripts" property missing from package.json
- `p-s` version: 3.0.3
- `node` version: 7.4.0
- `npm` version: 4.0.5

```javascript
/p-s/dist/bin-utils/initialize/index.js
```

The command executed:

```console
./node_modules/.bin/p-s- init
```

The output:

```console
./node_modules/p-s/dist/bin-utils/initialize/index.js:31
   test: scripts.test ? 'nps test' : undefined
                 ^

TypeError: Cannot read property 'test' of undefined
```

Problem description:
Init fails to work if there is no "scripts" object in package.json

Suggested solution:
Inform user of problem or create "scripts" property automatically.

kentcdodds:
**What**: Closes #89

**Why**: Because we want to be able to init even if you have no
`scripts` in your `package.json`

**How**: ES6 magic

Thanks for the bug report @gunnx!


kentcdodds:
Anyone wanna review and merge this? It's pretty small.

Issue #100, opened 2017-02-09
Proper handling and forwarding of `--flags`


kentcdodds:
Right now you can do:

```console
nps foo --bar
```

And `p-s` will run the script called `foo` with the flag called `--bar`

However if you run:

```console
nps foo --silent --bar
```

then `p-s` will run the script called `foo` with the flag called `--bar` but without the flag called `--silent`. This is because of [this](https://github.com/kentcdodds/p-s/blob/3566a15b991d6b84e7386bd18b0e08c2e94c6235/src/bin-utils/index.js#L113-L124).

Normally it's not a huge deal. I think that's confusing.

Here are a few possible solutions (all of them are breaking):

1. Only accept `p-s` flags before the scripts (so you'd have to do `nps --silent foo --bar`)
2. Require `--` before the scripts and flags (`nps --silent -- foo --bar`)
3. Forward all flags whether they're `p-s` related or not (`nps foo --silent --bar` and forwards `--silent` and `--bar` both)

I can't think of a way to change this behavior without breaking things.

My favorite solution is the first. Mostly because it's least noisy and seems to make the most sense.

I'm working on #70 right now and will make this a part of that PR. I want feedback on this really quick. If anyone has any ideas let me know. If you'd like to vote for one or the other use github reactions to vote: :+1: for 1, :smile: for 2, :tada: for 3, and :confused: for "I don't know/care"

gunnx:
👍

kentcdodds:
Actually, yargs has a pretty sweet API for this. I think we'll do things in a way that's mostly backward compatible. If you want to forward flags into commands, you'll have to put those commands in quotes. This will make the most sense I think!

addityasingh:
@kentcdodds But do you think it would look a clean syntax?
```javascript
nps 'build --foo'
```

My 2 cents: I would still prefer to not have any special syntax for my cli commands and below would be better than above:
```javascript
nps build --foo
```

kentcdodds:
Yeah, I can see how the syntax looks much cleaner. As I described though, the problem is that there's no way to know whether the user intends for the command to be passed to `p-s` or `build`. Things get even more complicated when you have multiple scripts you're running too. Before we just forwarded all non-`p-s` flags on, but that got confusing if you wanted to pass on a flag to a script that `p-s` wouldn't forward (like `--silent`).

[...]

PR #101, opened 2017-02-09
fix(cli): improve handling of forwarding args/flags

### LINKED ISSUE #70: Move to yargs

kentcdodds:
**What**: This migrates us from `commander` and `omelette` to `yargs`
and in the process changes how we handle forwarding of arguments and
flags. We no longer just arbitrarily forward arguments to every script.
Also instead of comma separated scripts, they're simply separated by a
space. If you want to pass args/flags to a script, then you put it in
quotes.

**Why**: Makes this more predictable. Also `yargs` is more powerful in
general anyway.

**How**: Magic... Will comment inline...

Closes #70 and #100


kentcdodds:
Anyone watching the repo can probably tell that I'm at a total loss on why the build is failing. Anyone wanna give it a look and figure out what's up? I think somehow that `p-s` is spawning processes in an old version of node... Or something. Totally bonkers.

kentcdodds:
I'm fairly confident that I've narrowed it down to `spawn-default-shell` (dependency of `concurrently`) being the problem (cc @kimmobrunfeldt). My guess is that the way that it determines the shell to use ignores `nvm` which Travis has set up for the correct node version and it uses the pre-existing node version. I'm working on [a repro repo](https://github.com/kentcdodds/spawn-default-shell-node-version-bug) right now.

kentcdodds:
[Verified](https://github.com/kimmobrunfeldt/spawn-default-shell/issues/8)

kentcdodds:
Looks like my fork fixes the issue. Just waiting on [this](https://github.com/kimmobrunfeldt/concurrently/pull/89) to be merged and released or finding another solution.

kentcdodds:
Wahoo! This PR is ready for review finally!

src/index.test.js
```diff
@@ -60,24 +60,15 @@ test('options: logLevel sets the log level', () => {
   })
 })
 
-test('passes on additional arguments', () => {
```
kentcdodds: We don't worry about this anymore. This is no longer a feature. If you want to pass arguments, put it in quotes. It's less confusing this way

addityasingh:
@kentcdodds The changes look fine, but do you think it would look a clean syntax?
```javascript
nps 'build --foo'
```

My 2 cents: I would still prefer to not have any special syntax for my cli commands and below would be better than above:
```javascript
nps build --foo
```

kentcdodds:
Thanks for the review @addityasingh (I've added you to the contributors table as a reviewer!),
I'll respond to you on #100

src/get-logger.js
```diff
@@ -50,6 +52,16 @@ function getShouldLogFn(...acceptableValues) {
   acceptableValues = ['', 'debug', ...acceptableValues]
```
addityasingh: I think it's preferable to treat params (`acceptableValues `) as const and not to override them, to avoid any kind of confusion in readability

src/get-logger.js
```diff
@@ -50,6 +52,16 @@ function getShouldLogFn(...acceptableValues) {
...
   return function shouldLogWithLevel(logLevel = '') {
     logLevel = logLevel.toLowerCase()
-    return !logLevel || acceptableValues.some(v => logLevel === v)
+    return !logLevel || includes(acceptableValues, logLevel)
+  }
+}
+
+function getLogLevel({silent, logLevel}) {
```
addityasingh: This whole function can be refactored as below:
```javascript
const getLogLevel = ({silent, logLevel}) => silent ? 'disable' : logLevel
```

kentcdodds:
I'm going to merge this now and publish a beta for the `next` branch. I think there's still more to do, but I want to get feedback!

gunnx:
I just updated and realised I can no longer pass down args as I was doing before:
e.g.
`npm start lint` - run linter
`npm start lint -- --fix` - run linter with fix flag passed

I guess this was just useful sometimes without having to create every possible option in package-scripts

kentcdodds:
As noted in [the release notes](https://github.com/kentcdodds/nps/releases/tag/v5.0.0)

> **Another significant change here** is the change for #100.
> 
> **before**:
> 
> ```
> nps lint,build "src/**/*.js"
> ```
> 
> **after**:
> 
> ```
> nps "lint src/**/*.js" "build src/**/*.js"
> ```
> 
> It's kinda sad, but I honestly never really saw anyone utilizing the argument forwarding feature anyway. Where this gets a little more impactful is when you're typing at the command line:
> 
> **before**:
> 
> ```
> nps test --updateSnapshot
> ```
> 
> **after**:
> 
> ```
> nps "test --updateSnapshot"
> ```
> 
> Notice the quotes. Those are now required for a command and its arguments to be grouped (otherwise `--updateSnapshot` will be assumed to be a command you're trying to execute and you'll get an error). Actually, `nps` will throw an error if you try `nps foo --flag` because `--flag` is not a recognized `nps` flag :)

:)

Issue #82, opened 2016-12-20
Set up all npm environment variables


kentcdodds:
If you run `npm run env` it'll show you all the environment variables that npm creates for you. It'd be handy to include those in the environment for the script that's run so there are fewer surprises when migrating from npm scripts to p-s, also, it can be handy in general :)

sloansparger:
@kentcdodds, I'm up for taking this one. Do you mind helping me out a little with the direction you would like it to go?

I see you have a getEnv function, would you like for me to handle this functionality there or pull in a function from another file that returns these env variables and merge the two in getEnv?

kentcdodds:
Cool @sloansparger! Yes, I think this should happen in `getEnv` and I'm thinking that we can move [it](https://github.com/kentcdodds/nps/blob/master/src/index.js#L84-L92) to its own module so it can be tested in isolation (so kind of a combination of what you suggested). I'm not certain how complicated it will be so it may be that you'll need to have a function or two in the new module and only expose the one `getEnv` as the default export. Thanks for taking this on! Please make sure to follow [the contributing guidelines](https://github.com/kentcdodds/nps/blob/master/CONTRIBUTING.md). :)

v1adko:
@kentcdodds As I understand this just opens/proxies npm environment variables to the scripts run via nps, right? 
What about setting your own environment variables?
I use [better-npm-run](https://www.npmjs.com/package/better-npm-run) (which, in fact, depend are based on your cross-env) to provide variables declaratively (useful for swapping database connection parameters and other API tokens for dev/test). 
But 'bnr' bloats the package.json even more, this could really help avoid that.

kentcdodds:
> What about setting your own environment variables?

I recommend just using `cross-env`. You can even use [`nps-utils`](https://github.com/kentcdodds/nps-utils) which has a utility function for that :+1:

PR #121, opened 2017-03-01
WIP want to make sure this is what we actually want

sloansparger:
#82

@kentcdodds, I was able to complete this feature, but wanted to run by you if the output was what you wanted. It seems like the running npm run env sets some global variables that wouldn't be the same if you ran another script. Running npm run test should highlight in red pretty clearly what I'm talking about.

If we want to move forward, I still lack cleaning this up a little. I would super appreciate some input on how you want to handle errors and direction on how to test these functions well.

### DIFF
--- a/src/get-env-variables.js
+++ b/src/get-env-variables.js
@@ -1,0 +1,60 @@
+const exec = require('child_process').exec
+// import spawn from 'spawn-command-with-kill'
+import {assign, clone} from 'lodash'
+import managePath from 'manage-path'
+import {sync as findUpSync} from 'find-up'
+
+const NON_ERROR = 0
+
+export default getEnv
+
+function getEnv() {
+  return new Promise((resolve, reject) => {
+    getNpmEnvVariables()
+      .then(npmVars => {
+        const env = assign(clone(process.env), npmVars)
+        const alterPath = managePath(env)
+        const npmBin = findUpSync('node_modules/.bin')
+        if (npmBin) {
+          alterPath.unshift(npmBin)
+        }
+        resolve(env)
+      })
+      .catch(reject)
+  })
+}
+
+function getNpmEnvVariables() {
+  let child
+  return new Promise((resolve, reject) => {
+    let output = ''
+    child = exec('npm run env')
+
+    child.stdout.on('data', chunk => output += chunk)
+
+    // NOTE: not sure what needs to happen with error logging
+    child.stderr.on('data', data => console.log(`stderr: ${data}`))
+
+    child.on('close', code => {
+      if (code === NON_ERROR) {
+        let npmVariables = parseNpmRunEnvOutput(output)
+        resolve(npmVariables)
+      } else {
+        // NOTE: again, not sure how we want to handle errors being logged
+        reject('npm run env exited with status code:', code)
+      }
+    })
+  })
+}
+
+function parseNpmRunEnvOutput (output) {
+  return output
+    .trim()
+    .split('\n')
+    .slice(3)
+    .map(line => line.split('='))
+    .reduce((acc, [key, val]) => {
+      acc[key] = val
+      return acc
+    }, {})
+}
--- a/src/index.js
+++ b/src/index.js
@@ -1,13 +1,12 @@
 import spawn from 'spawn-command-with-kill'
 import chalk from 'chalk'
 import {oneLine} from 'common-tags'
-import {isString, clone} from 'lodash'
-import {sync as findUpSync} from 'find-up'
-import managePath from 'manage-path'
+import {isString} from 'lodash'
 import arrify from 'arrify'
 import getScriptToRun from './get-script-to-run'
 import getScriptsFromConfig from './get-scripts-from-config'
 import getLogger, {getLogLevel} from './get-logger'
+import getEnv from './get-env-variables'
 
 const NON_ERROR = 0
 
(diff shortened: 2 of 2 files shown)


src/get-env-variables.js
```diff
@@ -0,0 +1,60 @@
...
+import {assign, clone} from 'lodash'
+import managePath from 'manage-path'
+import {sync as findUpSync} from 'find-up'
+
+const NON_ERROR = 0
+
+export default getEnv
+
+function getEnv() {
+  return new Promise((resolve, reject) => {
```
kentcdodds: You can rewrite this by simply chaining the promise. You don't need to use the promise constructor, you can simply do:

```javascript
return getNpmEnvVariables().then(npmVars => {
  const env = assign(clone(process.env), npmVars)
  const alterPath = managePath(env)
  const npmBin = findUpSync('node_modules/.bin')
  if (npmBin) {
    alterPath.unshift(npmBin)
  }
  return env
})
```

src/get-env-variables.js
```diff
@@ -0,0 +1,60 @@
...
+        resolve(npmVariables)
+      } else {
+        // NOTE: again, not sure how we want to handle errors being logged
+        reject('npm run env exited with status code:', code)
+      }
+    })
+  })
+}
+
+function parseNpmRunEnvOutput (output) {
```
kentcdodds: Please remove the space between the function name and the `(`.

src/get-env-variables.js
```diff
@@ -0,0 +1,60 @@
...
+      })
+      .catch(reject)
+  })
+}
+
+function getNpmEnvVariables() {
+  let child
+  return new Promise((resolve, reject) => {
+    let output = ''
+    child = exec('npm run env')
```
kentcdodds: Is this the only way to get the environment variables? Could you perhaps check npm's code to see whether they expose this via a node API? Would be really nice to not have to spawn anything...

src/get-env-variables.js
```diff
@@ -0,0 +1,60 @@
+const exec = require('child_process').exec
+// import spawn from 'spawn-command-with-kill'
```
kentcdodds: I'd prefer to use `spawn` if we can.

sloansparger: Ah, totally forgot to mention this in my comment! Spawn wasn't working... I pulled it out into a separate file and ran it and everything works just fine. Do you know why that might be?

src/get-env-variables.js
sloansparger: Let me take a look and see what I find.

src/get-env-variables.js
kentcdodds: Huh, that's odd. No idea... 🤔

src/get-env-variables.js
sloansparger: Don't know if we can avoid having to spawn anything.

The only promising thing that I could find was this: https://github.com/npm/npm/blob/d081cc6c8d73f2aa698aab36605377c95e916224/lib/config.js#L188. But this would require us to have npm as a dependency, which doesn't sound like a better option to me.

src/get-env-variables.js
sloansparger: Okay, another 2 hours of trying to get it to work and I have been left frustrated...

When I run inside of this repo:
let child = spawn('npm run env')
child.stdout.on('data', () => {})
// throws error can't read on of undefined

When I run outside of this repo, everything works just fine...

If you have time to help that would be greatly appreciated, otherwise I may not be any use for moving forward on this :(

kentcdodds: Let's just go with what you have :+1:

src/get-env-variables.js
kentcdodds: That's fine. Let's just `exec`. No problem.

kentcdodds:
Thanks so much for working on this. If you could, please [rebase your changes](https://egghead.io/lessons/javascript-how-to-rebase-a-git-pull-request-branch) and reinstall dependencies. I've made a few updates that you should probably get. Thanks again!

src/get-env-variables.js
v1adko: @kentcdodds What about adding prettier to this project? Would help avoid all of the nits and potentially save your time :)

kentcdodds: Yeah, eventually I'll upgrade this repo to the latest stuff I have for my repos...

v1adko: Do you have a boilerplate for the basic OSS projects? Common things like build scripts + testing (I assume Jest for both node and react) + lint and prettier on precommit, CI config... Possibly documentation/guides for bots like LGTM or automatic issue expiration. Maybe some other cool things I don't know about that are common for OSS :)
I imagine it would be easier to base all of the new OSS and then update them later on via an upstream. Or you do the projects usually differ too much to abstract all of the common things.
Would be cool to have some kind of 'create-oss-app' I think. Or maybe there are some that I don't know about?

kentcdodds: I have [a boilerplate for my projects](https://github.com/kentcdodds/generator-kcd-oss). I've often considered making `create-oss-lib`, but I've not taken the time because I find that often libs vary in needs pretty widely. May be able to get away with `create-react-lib` and `create-node-cli` or `create-js-lib`, but trying to make them all work in one package would be difficult.

PR #122, opened 2017-03-02
fix: add nps-utils to docs and package-scripts.js

kentcdodds:
**What**: This adds `nps-utils` to the docs and `package-scripts.js`

**Why**: Because people miss the simplicity of `--parallel` among other things :)

**How**: mostly docs changes, but some of the changes are in `package-scripts.js`

You might like this @tleunen :) And you should probably know about this @elijahmanor

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ All the benefits of npm scripts without the cost of a bloated package.json and l
 ## Quick Video Intro :tv:
 
 <a href="http://kcd.im/nps-video" title="Pull out npm scripts into another file with nps">
-  <img src="other/video-screenshot.png" alt="Video Screenshot" title="Video Screenshot" width="700" />
+  <img src="https://github.com/kentcdodds/nps/raw/master/other/video-screenshot.png" alt="Video Screenshot" title="Video Screenshot" width="700" />
 </a>
 
 [Pull out npm scripts into another file with nps][video] by [Elijah Manor](https://github.com/elijahmanor) (5:53)
@@ -41,6 +41,8 @@ this file is a JavaScript file, you can do a lot more with your project scripts.
 `package-scripts.js` file:
 
 ```javascript
+const npsUtils = require('nps-utils') // not required, but handy!
+
 module.exports = {
   scripts: {
     default: 'node index.js',
@@ -58,9 +60,8 @@ module.exports = {
       default: 'webpack',
       prod: 'webpack -p',
     },
-    // learn more about concurrently here: https://npm.im/concurrently
-    validate: 'concurrently "nps lint" "nps test" "nps build"',
-    // concurrently script too verbose for your liking? Check out other/EXAMPLES.md!
+    // learn more about npsUtils here: https://npm.im/nps-utils
+    validate: npsUtils.concurrently.nps('lint', 'test', 'build'),
   },
 }
 ```
@@ -410,6 +411,11 @@ This was inspired by [a tweet][tweet] by [@sindresorhus][sindre].
 Big thank you to [@tmpvar][tmpvar] for giving up the name `nps`! The original `nps` is now
 called [`npmsearch-cli`](https://www.npmjs.com/package/npmsearch-cli).
 
+## Related Packages
+
+- [`nps-utils`][nps-utils] - a collection of utilities to make cross-platform scripts and many other patterns
+(like running concurrent/parallel scripts)
+
 ## Other Solutions
 
 - [scripty][scripty] has a solution for this problem as well. The reason I didn't go with that though is you still need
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ All the benefits of npm scripts without the cost of a bloated package.json and l
 ## Quick Video Intro :tv:
 
 <a href="http://kcd.im/nps-video" title="Pull out npm scripts into another file with nps">
-  <img src="other/video-screenshot.png" alt="Video Screenshot" title="Video Screenshot" width="700" />
+  <img src="https://github.com/kentcdodds/nps/raw/master/other/video-screenshot.png" alt="Video Screenshot" title="Video Screenshot" width="700" />
```
kentcdodds: This way the screenshot shows up on the npm registry listing :)

README.md
```diff
@@ -41,6 +41,8 @@ this file is a JavaScript file, you can do a lot more with your project scripts.
 `package-scripts.js` file:
 
 ```javascript
+const npsUtils = require('nps-utils') // not required, but handy!
```
kentcdodds: Want to show it early on.

README.md
```diff
@@ -58,9 +60,8 @@ module.exports = {
       default: 'webpack',
       prod: 'webpack -p',
     },
-    // learn more about concurrently here: https://npm.im/concurrently
-    validate: 'concurrently "nps lint" "nps test" "nps build"',
-    // concurrently script too verbose for your liking? Check out other/EXAMPLES.md!
+    // learn more about npsUtils here: https://npm.im/nps-utils
+    validate: npsUtils.concurrently.nps('lint', 'test', 'build'),
```
kentcdodds: Funny that it's actually a little longer than before, but I think it's more straightforward :man_shrugging:

other/EXAMPLES.md
```diff
@@ -53,26 +59,31 @@ get the idea 😄. This is a pretty nice win over traditional npm scripts 👍
 ### parallel scripts
 
 Often, scripts can run concurrently because they are not interdependent. We recommend
-[`concurrently`](http://npm.im/concurrently) for this:
+[`nps-utils`](http://npm.im/nps-utils) which uses `concurrently` for this:
 
 ```javascript
+const npsUtils = require('nps-utils')
```
kentcdodds: Making this example more complete.

package-scripts.js
```diff
@@ -42,22 +44,8 @@ module.exports = {
...
-      script: concurrent({
-        'build-and-validate': {
-          script: 'nps build.andValidate',
-          color: 'bgBlue.bold',
-        },
-        test: {
-          script: 'nps test',
-          color: 'bgMagenta.bold',
-        },
-        'lint-staged': {
```
kentcdodds: `lint-staged` shouldn't happen during the `validate` script. Only during `precommit`, so I've moved it to that.

package-scripts.js
```diff
@@ -42,22 +44,8 @@ module.exports = {
...
-          script: 'nps lintStaged',
-          color: 'bgGreen.bold',
-        },
-      }),
+      script: concurrent.nps('test', 'build.andValidate'),
```
kentcdodds: How's that for simpler? :)

package-scripts.js
```diff
@@ -76,35 +64,6 @@ module.exports = {
   },
 }
 
-function concurrent(scripts) {
```
kentcdodds: Yay!

package-scripts.js
```diff
@@ -76,35 +64,6 @@ module.exports = {
...
-    colors,
-    scripts: quotedScripts,
-    names,
-  } = Object.keys(scripts).reduce(reduceScripts, {
-    colors: [],
-    scripts: [],
-    names: [],
-  })
-  const flags = [
-    //   https://github.com/kimmobrunfeldt/concurrently/issues/91
```
kentcdodds: This was fixed with a new flag `--kill-others-on-fail`, but you don't have to worry about that, it's all taken care of for you. In fact, you don't even need to install `concurrently` yourself! Just install `nps-utils` and it'll magically work :)

package.json
```diff
@@ -54,9 +52,9 @@
     "jest-cli": "^18.1.0",
     "lint-staged": "^3.3.0",
     "nps": "*",
+    "nps-utils": "^1.1.0",
     "opt-cli": "^1.5.1",
     "prettier-eslint-cli": "^3.0.0",
-    "rimraf": "^2.6.0",
```
kentcdodds: We don't need this anymore because `nps-utils` takes care of it for us (it actually still uses `rimraf`, but it brings in the dep for us).

package.json
```diff
@@ -42,10 +42,8 @@
     "babel-preset-env": "^1.1.8",
     "babel-preset-stage-2": "^6.22.0",
     "babel-register": "^6.23.0",
-    "cli-tester": "^2.0.0",
```
kentcdodds: We haven't used this for a while, removing...

package.json
```diff
@@ -42,10 +42,8 @@
...
     "codecov": "^1.0.1",
     "commitizen": "^2.9.6",
-    "concurrently": "^3.3.0",
```
kentcdodds: `nps-utils` brings this in for us and manages finding the right binary for us :)

Issue #124, opened 2017-03-08
init command fails with a null path TypeError


wopian:
- `nps` version: 5.0.4
- `node` version: 7.2.1
- `npm` version: 4.1.2

`package.json`:

```json
"scripts": {
    "dev": "node build/dev-server.js",
    "build": "node build/build.js",
    "test": "cross-env BABEL_ENV=test karma start test/unit/karma.conf.js --single-run",
    "lint": "eslint --ext .js,.vue client test/unit/specs"
  },
```

The command executed:

```console
./node_modules/.bin/nps init --type yml
```

The output:

```console
path.js:7
    throw new TypeError('Path must be a string. Received ' + inspect(path));
    ^

TypeError: Path must be a string. Received null
    at assertPath (path.js:7:11)
    at Object.resolve (path.js:186:7)
    at Function.module.exports.sync (C:\github\hibari\node_modules\locate-path\index.js:20:28)
    at Function.module.exports.sync (C:\github\hibari\node_modules\nps\node_modules\find-up\index.js:38:27)
    at Object.onInit [as handler] (C:\github\hibari\node_modules\nps\dist\bin-utils\parser.js:155:49)
    at Object.self.runCommand (C:\github\hibari\node_modules\nps\node_modules\yargs\lib\command.js:170:22)
    at parseArgs (C:\github\hibari\node_modules\nps\node_modules\yargs\yargs.js:920:28)
    at Object.Yargs.self.parse (C:\github\hibari\node_modules\nps\node_modules\yargs\yargs.js:499:18)
    at parse (C:\github\hibari\node_modules\nps\dist\bin-utils\parser.js:85:27)
    at Object.<anonymous> (C:\github\hibari\node_modules\nps\dist\bin\nps.js:22:33)
```

Problem description:

Running the initialisation script to generate and update package.json automatically fails with the above error with the 4 scripts I currently have.

[Repo](https://github.com/wopian/hibari)

kentcdodds:
Thanks for the report. I'll get to it when I can. Anyone wanna dig in a bit?

wopian:
Further note: windows 10 and with VSCode 1.10's terminal.

Same scripts written in `package-scripts.yml` function just fine with `"start": "nps"` in `package.json`:

```yml
scripts:
  dev:
    script: node build/dev-server.js
    description: run server and watch for changes
  build:
    default:
      script: node build/build.js
      description: build from source code
    report:
      script: node build/build.js --report
[...]

PR #127, opened 2017-03-21
Pr/fix init error

natecavanaugh:
When findUp can't find an existing package-scripts.js file, it returns null, which is passed back to
findUp and throws an error because path.resolve can only accept strings as arguments

#124

**What**: Fixes error thrown when a package-scripts.js file doesn't exist while running `init`

**Why**: Because if you're running `nps init` without an existing package-scripts.js file, it won't work, forcing developers to have to create the file manually.

**How**: Checking that the `path` returned from `getPSConfigFilepath` is actually a string before passing it to `findUp.sync`.

I tried figuring out why the tests never fail for this, but I believe it's because of jest's mocking of findUp means that it never even hits the `path.resolve` method, so I'm not sure how you'd want to test for this particular case.


.all-contributorsrc
```diff
@@ -266,6 +266,13 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "natecavanaugh",
+      "name": "Nate Cavanaugh",
+      "avatar_url": "https://avatars1.githubusercontent.com/u/116871?v=3",
+      "profile": "http://alterform.com",
+      "contributions": []
```
kentcdodds: Could you add `"code"` to this array and then run `npm start generateContributors`?

kentcdodds:
This is great, thank you!

.all-contributorsrc
natecavanaugh: Done, thanks for the heads up :)

kentcdodds:
Thank you so much!

Issue #129, opened 2017-04-14
Test fails under Windows OS


Miklet:
Hello, I wanted to start contributing to this project, but unfortunately tests are failing under Windows 8/10 and probably on any other Windows OS too.

- `nps` version: 5.0.5
- `node` version: 6.9.2
- `npm` version: 3.9.6

Scripts file (or at least the relevant bits):

```javascript
src\bin-utils\index.test.js
src\get-scripts-from-config.test.js
```

The command executed:

```console
npm start validate
```

The output:

```console
 ● loadConfig: logs and throws an error for a config that exports the wrong data type
[test]
[test]     expect(jest.fn()).toHaveBeenCalledWith(expected)
[test]
[test]     Expected mock function to have been called with:
[test]       [{"message": {"regexp": /C:\code\github\nps\src\bin-utils\fixtures\bad-data-type-config.js/}, "ref": "config-must-be-an-object"}]
[test]     But it was called with:
[test]       [{"message": "The package-scripts configuration (\"C:\\code\\github\\nps\\src\\bin-utils\\fixtures\\bad-data-type-config.js\") must be an object or a function that returns an object.", "ref": "config-must-be-an-object"}]
[test]
[test]       at Object.<anonymous> (src\bin-utils\index.test.js:65:21)
[test]
[test]   ● loadConfig: logs and throws an error for a config that exports a function that returns the wrong data type
[test]
[test]     expect(jest.fn()).toHaveBeenCalledWith(expected)
[test]
[test]     Expected mock function to have been called with:
[test]       [{"message": {"regexp": /C:\code\github\nps\src\bin-utils\fixtures\bad-function-data-type-config.js/}, "ref": "config-must-be-an-object"}]
[test]     But it was called with:
[test]       [{"message": "The package-scripts configuration (\"C:\\code\\github\\nps\\src\\bin-utils\\fixtures\\bad-function-data-type-config.js\") must be an object or a function that returns an object.", "ref": "config-must-be-an-object"}]
[test]
[test]       at Object.<anonymous> (src\bin-utils\index.test.js:90:21)
```
```
FAIL  cli-test\cli.test.js
[build.andValidate]   ● with config with default script
[build.andValidate]
[build.andValidate]     Error
[build.andValidate]       Error: expect(value).toMatchSnapshot()
[build.andValidate]
[build.andValidate]       Received value does not match stored snapshot 1.
[build.andValidate]
[build.andValidate]       - Snapshot
[build.andValidate]       + Received
[...]

PR #132, opened 2017-04-15
test(cli): fixed tests under windows

Miklet:
I've fixed tests for cli under windows os. Problems with tests were caused because of platform
diffreneces between windows and linux i.e. paths separator, output from echo command.

https://github.com/kentcdodds/nps/issues/129

**What**: I've changed tests for cli to run them successfully under windows.

**Why**: Because people like I, who use windows, couldn't setup project locally properly because of tests failing on our operating system.

**How**: I have tweaked one method and added a new one which removes unnecessary quotes from stdout result in tests output.

P.S. forgive me my lack of knowledge about contributing, it is my first time.


Miklet:
I ran commands and there are no changes :) I've just adjusted tests too be more versatile and still match actual output without changing it.

kentcdodds:
Oh, awesome! Thank you again!

kentcdodds:
Congrats on your first pull request! Thank you for your contribution!

Miklet:
Thank you very much, I feel great that I could help. I saw that there are still some issues to be done, will look at them to help a little more.

kentcdodds:
Terrific!

Issue #126, opened 2017-03-15
Add script name in the `nps is executing` output


kentcdodds:
Right now if I execute:

```
nps f.b
```

I'll get something like:

```
nps executing: echo "foo bar"
```

I'd like it to say:

```
nps is executing foo.bar: echo "foo bar"
```

where `foo.bar` is the name (path) of the script being executed. It'd be good to get it in bold too.

Anyone wanna take this up?

Miklet:
Since my PR for `prefix-matches` got merged and now it returns corresponding property chain I can get back to this issue. I have question about module `resolve-script-object-to-string` if it is still needed after changes. Result of `prefix-matches` will now always return an object and also we want now to get resolved prefix with script to execute, not only script.

kentcdodds:
Yes, the PR to support this will be much simpler now :)

Miklet:
If it will be much simpler then I guess we can drop module `resolve-script-object-to-string` as I asked earlier :)?

kentcdodds:
I expect so :) Thanks so much!

PR #136, opened 2017-04-23
refactor(cli): display which script is executed

Miklet:
Added an additional information displaying which script is executed.

**What**: Added information which script is executed.

**Why**: Because of #126 

**How**: Just added an additional chalk call and changed string concatenation for template literal.

### DIFF
--- a/cli-test/__snapshots__/cli.test.js.snap
+++ b/cli-test/__snapshots__/cli.test.js.snap
@@ -12,7 +12,7 @@ lint.sub.hiddenThing
 exports[`test with --require 1`] = `
 Object {
   "stderr": "",
-  "stdout": "nps executing: echo \"log\"
+  "stdout": "nps is executing \`log\`: echo \"log\"
 log
 ",
 }
@@ -37,8 +37,17 @@ Object {
 exports[`test with config with default script 1`] = `
 Object {
   "stderr": "",
-  "stdout": "nps executing: echo \"default script\"
+  "stdout": "nps is executing \`default\`: echo \"default script\"
 default script
+",
+}
+`;
+
+exports[`test with prefix 1`] = `
+Object {
+  "stderr": "",
+  "stdout": "nps is executing \`lint.sub.thing.script\`: echo \"deeply nested thing\"
+deeply nested thing
 ",
 }
 `;
--- a/cli-test/cli.test.js
+++ b/cli-test/cli.test.js
@@ -17,6 +17,9 @@ test('with --require', () =>
 test('with --get-yargs-completions', () =>
   snapshot('--config ./package-scripts.js --get-yargs-completions li'))
 
+test('with prefix', () =>
+  snapshot('--config ./package-scripts.js lint.s.t.s'))
+
 function snapshot(args) {
   return runNPS(fixturesPath, args).then(results => {
     const snapshottableResults = convertResultToLinuxSpecific(results)
--- a/src/get-script-to-run.js
+++ b/src/get-script-to-run.js
@@ -1,4 +1,4 @@
-import {each, cloneDeep, isPlainObject} from 'lodash'
+import {each, cloneDeep, isPlainObject, isUndefined, isString} from 'lodash'
 import prefixMatches from 'prefix-matches'
 import resolveScriptObjectToString from './resolve-script-object-to-string'
 import kebabAndCamelCasify from './kebab-and-camel-casify'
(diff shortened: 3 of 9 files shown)


kentcdodds:
Thanks for working on this!

Miklet:
It looks like that:
test

kentcdodds:
Thanks for the screenshot! The use case is: I've entered a prefix version of the script, so I know what I entered. I want to see the full name of the script to validate that the script that's running is the one I'm thinking. Does that make sense?

Miklet:
Yes, it does make sense if I understood it properly 😄 So instead of:
`nps is executing fo.b.ba: echo "foo.bar.baz"`
you expect:
`nps is executing foo.bar.baz: echo "foo.bar.baz"`. Am I right?

kentcdodds:
Correct 😀

Miklet:
It seemed to be pretty easy issue but it turned out to be a little bit more complicated. I thought that there is somewhere some lonely variable with resolved script name but I was wrong 😄 I am thinking now about some function which will try to find matching key by a value by traversing recurisvely script's object. I will encapsulate it to separate module and it will be easy to test.

Miklet:
Added script's name resolving. Let me know if you like it and I will squash commits into one before merge.

src/resolve-script-name.js
```diff
@@ -0,0 +1,28 @@
+import {get} from 'lodash'
+
+export default resolveScriptName
+
+function resolveScriptName(scriptToFind, scriptsConfig) {
```
kentcdodds: Rather than `scriptToFind`, could this be passed the prefix? (like `fo.baz.f` could find `foo.baz.foobar`)

Miklet: Not really, because this function search for a property which contains given value (`scriptToFind`). Function which resolve what you mentioned is already implemented and it is a `prefix-matcher` package.

kentcdodds: Sorry, what I meant to say was, that's what I'd _rather_ it do :smile:

Miklet: I understand, but I actually think the way it is, it is better. Because I don't see a way to get resolved script name by his prefix in other way than reimplementing `prefix-matcher` 😄 I may be wrong, correct me if you see other solution.

tleunen: Couldn't we extend `getScriptToRun` to get the script name in addition to the script command?

Miklet: You mean instead of creating new one?

kentcdodds: > Couldn't we extend getScriptToRun to get the script name in addition to the script command?

Maybe. Without looking at the code I'm unsure. If so, that'd be great.

> reimplementing `prefix-matcher`

Maybe we could contribute back and expose a method to do what we need :)

kentcdodds: Note that I have merge (and maybe release?) rights on that repo.

Miklet: The best solution in my opinion would to be contribute to `prefix-matcher`, implement returning property (whole chain, not only the deepest) and then we could extend `getScriptToRun` with it.

kentcdodds: That sounds good to me :+1:

Miklet:
Hi, I bumped up a version of `prefix-matches` and used the newest version to resolve the issue.

cli-test/__snapshots__/cli.test.js.snap
```diff
@@ -12,7 +12,7 @@ lint.sub.hiddenThing
 exports[`test with --require 1`] = `
 Object {
   "stderr": "",
-  "stdout": "nps executing: echo \"log\"
+  "stdout": "nps is executing log: echo \"log\"
```
kentcdodds: Could this be: 

```
nps is executing `log`: echo "log"
```

cli-test/cli.test.js
```diff
@@ -17,6 +17,9 @@ test('with --require', () =>
 test('with --get-yargs-completions', () =>
   snapshot('--config ./package-scripts.js --get-yargs-completions li'))
 
+test('with prefix', () =>
```
kentcdodds: Cool :+1:

src/get-script-to-run.js
```diff
@@ -10,19 +10,30 @@ function getScriptToRun(config, input) {
...
   const [script] = prefixMatches(input, config)
-  return resolveScriptObjectToString(script)
+  if (script) {
+    const resolvedPrefix = Object.keys(script).shift()
+    let scriptToRun = script[resolvedPrefix]
+    if (resolvedPrefix && isPlainObject(scriptToRun)) {
+      scriptToRun = resolveScriptObjectToString(scriptToRun)
+    }
+    return {
+      resolvedPrefix,
```
kentcdodds: Could we call this `scriptName` instead? Because it's not the prefix, it's actually what the prefix was resolved to, which should be the full script name.

Miklet: Yea, that would actually be much better name.

Miklet:
Updated with your comments.

kentcdodds:
Thank you for this!

PR #146, opened 2017-06-15
Add support to display help for one specific script

imbhargav5:
**What**

Support help for specific scripts. "npm help build" will now log the script info for the script "build".

** Why **
Fixes issue #61 

** How**
1. Extracted a function `stringifyScript` out of help which renders one script as a message.
2. Used the `stringifyScript` function within a new function `specificHelpScript` to display help information for a script instead of `help` in said scenario.
3. Wrote tests to cover the scenario.

### DIFF
--- a/src/bin-utils/index.js
+++ b/src/bin-utils/index.js
@@ -5,7 +5,9 @@ import {
   isPlainObject,
   isUndefined,
   isEmpty,
+  isEqual,
   isFunction,
+  find,
 } from 'lodash'
 import typeOf from 'type-detect'
 import chalk from 'chalk'
@@ -99,7 +101,15 @@ function loadConfig(configPath, input) {
   return config
 }
 
-export {initialize, help, getModuleRequirePath, preloadModule, loadConfig}
+export {
+  initialize,
+  help,
+  getModuleRequirePath,
+  preloadModule,
+  loadConfig,
+  specificHelpScript,
+  stringifyScript,
+}
 
 /****** implementations ******/
 
@@ -161,22 +171,41 @@ function requireDefaultFromModule(modulePath) {
   }
 }
 
+function stringifyScript({name, description, script}) {
+  const coloredName = chalk.green(name)
+  const coloredScript = chalk.gray(script)
+  let line
+  if (description) {
+    line = [coloredName, chalk.white(description), coloredScript]
+  } else {
+    line = [coloredName, coloredScript]
+  }
+  return line.join(' - ').trim()
+}
+
+function specificHelpScript({scripts}, scriptName) {
+  const availableScripts = getAvailableScripts(scripts)
+  const script = find(
+    availableScripts,
+    availableScript =>
+      !availableScript.hiddenFromHelp &&
+      isEqual(availableScript.name, scriptName),
+  )
+  if (isPlainObject(script)) {
+    const scriptLines = stringifyScript(script)
+    const message = `${scriptLines}`
+    return message
+  } else {
+    return chalk.yellow(`Script ${scriptName} does not exist`)
+  }
+}
+
 function help({scripts}) {
   const availableScripts = getAvailableScripts(scripts)
   const filteredScripts = availableScripts.filter(
     script => !script.hiddenFromHelp,
   )
-  const scriptLines = filteredScripts.map(({name, description, script}) => {
-    const coloredName = chalk.green(name)
-    const coloredScript = chalk.gray(script)
-    let line
-    if (description) {
-      line = [coloredName, chalk.white(description), coloredScript]
-    } else {
-      line = [coloredName, coloredScript]
-    }
-    return line.join(' - ').trim()
-  })
+  const scriptLines = filteredScripts.map(stringifyScript)
   if (scriptLines.length) {
     const topMessage = 'Available scripts (camel or kebab case accepted)'
     const message = `${topMessage}\n\n${scriptLines.join('\n')}`
--- a/src/bin-utils/index.test.js
+++ b/src/bin-utils/index.test.js
@@ -3,7 +3,13 @@ import path from 'path'
 import chalk from 'chalk'
 import {spy} from 'sinon'
 import {oneLine} from 'common-tags'
-import {help, preloadModule, loadConfig} from './index'
+import {
+  help,
+  preloadModule,
+  loadConfig,
+  stringifyScript,
+  specificHelpScript,
+} from './index'
 
 test('preloadModule: resolves a relative path', () => {
   // this is relative to process.cwd() I think...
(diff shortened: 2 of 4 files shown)


bhargav175:
@kentcdodds Let me know what you think!

src/bin-utils/index.js
```diff
@@ -161,22 +171,41 @@ function requireDefaultFromModule(modulePath) {
...
+    line = [coloredName, chalk.white(description), coloredScript]
+  } else {
+    line = [coloredName, coloredScript]
+  }
+  return line.join(' - ').trim()
+}
+
+function specificHelpScript({scripts}, scriptName) {
+  const availableScripts = getAvailableScripts(scripts)
+  const script = find(
```
kentcdodds: Instead of using `find` here, let's use `prefixMatches` from `prefix-matches` (already installed). That way people can still use a prefix to get help on a script.

src/bin-utils/index.test.js
```diff
@@ -305,6 +311,69 @@ test(
...
+      },
+    }
+    const message = specificHelpScript(config, 'foo')
+    const expected = chalk.yellow('Script foo does not exist')
+    expect(message).toBe(expected)
+  },
+)
+
+test('specificHelpScript: formats a nice message', () => {
+  const config = {
```
kentcdodds: To make sure that you can handle nesting and prefixes, maybe change this config to:

```javascript
const config = {
  scripts: {
    foo: {
      bar: {
         description: 'the bar script',
          script: 'echo "bar"',
      }
    }
  }
}
```

And do:

```js
const message = specificHelpScript(config, 'fo.ba')
```

src/bin-utils/parser.js
```diff
@@ -110,8 +115,14 @@ function parse(rawArgv) {
     const noScriptSpecifiedAndNoDefault = !specifiedScripts.length &&
       !hasDefaultScript
     const hasHelpScript = Boolean(psConfig.scripts.help)
-    const commandIsHelp = isEqual(specifiedScripts, ['help']) && !hasHelpScript
-    if (commandIsHelp || noScriptSpecifiedAndNoDefault) {
+    const commandIsHelp = isEqual(specifiedScripts[0], 'help') &&
```
kentcdodds: This looks good :+1:

src/bin-utils/parser.test.js
```diff
@@ -205,6 +206,23 @@ test(
   },
 )
 
+test(
```
kentcdodds: Cool :+1:

kentcdodds:
Would you like to [add yourself](../blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) to the [contributors table](../#contributors)?

bhargav175:
Yes!  I will make the changes and also add myself as a contributor.  Will make changes in a bit.

src/bin-utils/index.js
bhargav175: How does prefix-matches help and work ? Also, I am not sure I fully understand what `prefix` means in this context. Can you explain ?

src/bin-utils/index.test.js
bhargav175: Yes. Makes sense. Nesting did work, but yes I should add an appropriate test case. Will do that.

kentcdodds: I'm pretty sure that the test case I've provided wouldn't work with the current implementation. Using `prefix-matches` should fix this. That's what prefix-matches does :)

bhargav175: Alright, when I use prefix-matches, I will be getting a nested object as the first member of the resultant Array. And the script can be a string, number or object. How would I validate it and convert it to 
```
{
    name ,
    description,
    script,
 }
```
format, so I can stringify it ?

kentcdodds: Hmmm... I'm not certain why the script could be a number. But if we look at the example from `prefix-match`'s README:

```javascript
const prefixMatches = require('prefix-matches')

// Resolves simple cases
prefixMatches('w', {
	watch: 'watch things',
	build: 'build things'
}) // => [{ watch: 'watch things' }]

// Does not flatten result set
prefixMatches('w', {
	watch: {
		js: 'watch javascript',
		css: 'watch css'
	},
	build: 'build things'
}) // => [{ watch: { js: 'watch javascript', css: 'watch css' }}]

// Resolves nested prefixes
prefixMatches('w.j', {
	watch: {
		js: 'watch javascript',
		css: 'watch css'
	},
	write: {
		js: 'write javascript'
	}
}) // => [{ 'watch.js': 'watch javascript' }, { 'write.js': 'write javascript' }]

// ... and so on
prefixMatches('b.f.j', {
	build: {
		frontend: {
			js: 'build javascript',
			css: 'build css'
		}
	}
}) // => [{ 'build.frontend.js': 'build javascript' }]
```

The `key` will be your `name`, and if the value is a string then that's your `script`, if it's an object then the `script` will be either the `default` or `script` value and the `description` will be the `description` value.

imbhargav5: Thanks for the nice explanation for prefix-matches. Makes sense. 
Let us assume this as config.
```
{
   scripts : {
       watch: {
		js: 'watch javascript',
		css: 'watch css'
	},
	build: 'build things'
   }
}
```
So from what I understand, in this case, `nps help w.j` and `nps help w.c` should return help definitions. But `nps help w` should not. (Even if prefix-matches has returned a non-empty array). Using prefix-matches will not be enough, IMO. I will also need to check if w.j is deterministic enough to resolve as a script. Am I missing something here?

Regarding script being a number, actually, I just wanted to gracefully handle invalid data (in case). There were some test cases with `script : 42`.

kentcdodds: Yeah, you know what, I think we should have a single function that takes a config and a prefix and resolves it to:

```
{
  name,
  script,
  description
}
```

That would be really helpful to use in other areas of the package as well. Do you think that you could write that?

imbhargav5: Yeah! That was my thought exactly. We should probably extract that behaviour from the code we already have. Shall I include it in this PR ?

kentcdodds: Hmmm... It would be easier to review if you made a separate PR to refactor things, then update this PR to use that.

Thank you so much for your work on this. It's incredibly valuable and helpful. This project needs a bit of love and you're giving it that. So thank you very much.

imbhargav5: Np. Alright. Sounds good!

kentcdodds: FYI, I'm going on a week-long family vacation starting tomorrow, so if I'm unresponsive for a while that's why :smile:

imbhargav5: Aah! Family time. Have a good one!

imbhargav5: Hey, I am almost done with the util. Can you just create an issue which describes what we are trying to achieve, so it's easier to ref ?

kentcdodds: Just reviewed your PR for the util. It's looking great. Just a few changes. Thank you so much!

imbhargav5: @kentcdodds I should probably refactor this PR.

kentcdodds: Haha, yeah, probably :smile:

erikfox:
@kentcdodds I still see this and #61 as open, does this still need to be finished?

kentcdodds:
Yeah, I think that @imbhargav5 is still working on this one.

imbhargav5:
@erikfox @kentcdodds I will update the PR this week.

imbhargav5:
Closing this PR in favor of #151

PR #148, opened 2017-06-20
feat: Add util to get a string based on prefix from a config

imbhargav5:
**What**:
This commit adds a util get-sript-by-prefix.js to bin/utils which takes in a config and a prefix
string as arguments and returns a script if it finds a match, or returns null

**Why**:
This util will help with the issue being fixed in #146 

**How**:
Added a util 'get-string-by-prefix.js' which takes in a config and prefix as arguments and returns a script object if found. Otherwise it returns null.

The script object has the following structure 
`{
   name,
  description,
  script
}`

### DIFF
--- a/src/bin-utils/get-script-by-prefix.js
+++ b/src/bin-utils/get-script-by-prefix.js
@@ -1,0 +1,60 @@
+import prefixMatches from 'prefix-matches'
+import {keys, isPlainObject, isString, has} from 'lodash'
+
+/*
+  Converts a string or object with a "script" key into an object with
+   {
+    name,
+    script,
+    description
+   }
+*/
+export function scriptToObject(name, scriptArg) {
+  if (isString(scriptArg)) {
+    return {
+      name,
+      script: scriptArg,
+      description: '',
+    }
+  }
+  if (isPlainObject(scriptArg)) {
+    if (has(scriptArg, 'default')) {
+      return scriptToObject(`${name}.default`, scriptArg.default)
+    } else {
+      const description = scriptArg.description || ''
+      const script = scriptArg.script
+      return {
+        name,
+        description,
+        script,
+      }
+    }
+  }
+  return null
+}
+
+export function isValidScript(script) {
+  if (isString(script)) {
+    return true
+  } else if (isPlainObject(script)) {
+    if (has(script, 'default')) {
+      return isValidScript(script.default)
+    }
+    return has(script, 'script')
+  } else {
+    return false
+  }
+}
+
+export default function getScriptByPrefix({scripts}, prefix) {
+  const matches = prefixMatches(prefix, scripts)
+  for (const match of matches) {
+    const matchedKeys = keys(match)
+    const name = matchedKeys[0]
+    const script = match[name]
+    if (isValidScript(script)) {
+      return scriptToObject(name, script)
+    }
+  }
+  return null
+}
--- a/src/bin-utils/get-script-by-prefix.js
+++ b/src/bin-utils/get-script-by-prefix.js
@@ -33,7 +33,7 @@ export function scriptToObject(name, scriptArg) {
   return null
 }
 
-export function isValidScript(script) {
+function isValidScript(script) {
   if (isString(script)) {
     return true
   } else if (isPlainObject(script)) {
@@ -48,13 +48,32 @@ export function isValidScript(script) {
 
 export default function getScriptByPrefix({scripts}, prefix) {
   const matches = prefixMatches(prefix, scripts)
+  // This array holds all the valid scripts in
+  // the order of priority (default scripts have lowest priority)
+  const matchedScriptsSortedByPriority = []
   for (const match of matches) {
     const matchedKeys = keys(match)
     const name = matchedKeys[0]
     const script = match[name]
     if (isValidScript(script)) {
-      return scriptToObject(name, script)
+      if (has(script, 'default')) {
+        // if it's a default script, push to the last of the array
+        matchedScriptsSortedByPriority.push({
+          name,
+          script,
+        })
+      } else {
+        // if it's not a default script, push to the first of the array
+        matchedScriptsSortedByPriority.unshift({
+          name,
+          script,
+        })
+      }
     }
+  }
+  if (matchedScriptsSortedByPriority.length) {
+    const {name, script} = matchedScriptsSortedByPriority[0]
+    return scriptToObject(name, script)
   }
   return null
 }
--- a/src/bin-utils/get-script-by-prefix.test.js
+++ b/src/bin-utils/get-script-by-prefix.test.js
@@ -265,19 +265,6 @@ describe(
       },
     )
 
-    test(
-      'Resolves a script when prefix can match default script as well',
-      () => {
-        const actual = getScriptByPrefix(config, 'f.b.d')
-        const expected = {
-          script: 'echo define',
-          name: 'foo.bar.define',
-          description: '',
-        }
-        expect(actual).toEqual(expected)
-      },
-    )
-
     test('Resolves default script when prefix def has no other match', () => {
       const actual = getScriptByPrefix(config, 'f.j.def')
       const expected = {
--- a/src/bin-utils/get-script-by-prefix.test.js
+++ b/src/bin-utils/get-script-by-prefix.test.js
@@ -1,0 +1,224 @@
+import {isPlainObject, isNull} from 'lodash'
+import getScriptByPrefix, {scriptToObject} from './get-script-by-prefix'
+
+test('scriptToObject returns null when script is not object or string ', () => {
+  const actual = scriptToObject('foo', 42)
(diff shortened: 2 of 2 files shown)


src/bin-utils/get-script-by-prefix.test.js
```diff
@@ -0,0 +1,224 @@
...
+  expect(isNull(actual)).toBeFalsy()
+})
+
+test('scriptToObject returns non-null object when script is string ', () => {
+  const scriptArg = {
+    script: 'hello',
+  }
+  const actual = scriptToObject('foo', scriptArg)
+  expect(isPlainObject(actual)).toBeTruthy()
+  expect(isNull(actual)).toBeFalsy()
```
kwelch: What is the structure of this object? From these tests it only shows that it is an object and not null.

src/bin-utils/get-script-by-prefix.test.js
```diff
@@ -0,0 +1,224 @@
...
+          script: 'echo foo.js',
+        },
+      },
+      foobar: {
+        description: 'Foobar script',
+        script: 'echo foobar',
+      },
+    },
+  }
+  const actual = getScriptByPrefix(config, 'f')
```
kwelch: I am not sure that I would expect this to be how it resolves conflicts. Will this always determine a "winner"?

src/bin-utils/get-script-by-prefix.js
```diff
@@ -0,0 +1,60 @@
...
+  if (isString(scriptArg)) {
+    return {
+      name,
+      script: scriptArg,
+      description: '',
+    }
+  }
+  if (isPlainObject(scriptArg)) {
+    if (has(scriptArg, 'default')) {
+      return scriptToObject(`${name}.default`, scriptArg.default)
```
kwelch: In is validScript, it tests before `default` and `script`. How is that covered here? I am not sure  am seeing it.

src/bin-utils/get-script-by-prefix.js
```diff
@@ -0,0 +1,60 @@
...
+  const matches = prefixMatches(prefix, scripts)
+  for (const match of matches) {
+    const matchedKeys = keys(match)
+    const name = matchedKeys[0]
+    const script = match[name]
+    if (isValidScript(script)) {
+      return scriptToObject(name, script)
+    }
+  }
+  return null
```
kwelch: I read through the other PR where it mentioned creating this utility. What I didn't see was how to break ties. I am not sure I am a fan of it resolving to which ever is first in the list.

kwelch:
The code looks good. I put in a few comments to help the code stand up on it owns. My biggest nit is how it resolves conflict. I think it should be possible for the scripts to have a tie, but I am not seeing that here. 

Additionally, if this main function for this current is for the help, as a user I think I would want to know about the nested scripts within the script name I passed. 

Just some thoughts.

src/bin-utils/get-script-by-prefix.test.js
imbhargav5: Ah, will add more tests to this one.

src/bin-utils/get-script-by-prefix.test.js
imbhargav5: It would find the first match. That is how prefixMatches works. So we just add more functions over it to get the script we want.

kwelch: I thought it returned an array of matches if multiple were close. I am fine with this, just want to ensure it is the expected functionality.

imbhargav5: Cool. Let's wait for @kentcdodds 's inputs, then? We will take it forward from there.

imbhargav5:
@kentcdodds ping!

src/bin-utils/get-script-by-prefix.test.js
```diff
@@ -0,0 +1,224 @@
...
+  }
+  const actual = scriptToObject('foo', scriptArg)
+  expect(isPlainObject(actual)).toBeTruthy()
+  expect(isNull(actual)).toBeFalsy()
+})
+
+test('getScriptByPrefix returns null when no matching script', () => {
+  const config = {scripts: {}}
+  const message = getScriptByPrefix(config, 'w')
+  expect(isNull(message)).toBeTruthy()
```
kentcdodds: all of these `isNull`s can be: `expect(message).toBeNull()`. We'll get better error messages :smile:

src/bin-utils/get-script-by-prefix.test.js
kentcdodds: This is correct. It currently finds the first match. Though, just to be clear. Currently if you provided `f` with these scripts, it would not match `foo.js`. You have to provide at least one character for each level of nesting. So this test is resolving correctly. :+1:

src/bin-utils/get-script-by-prefix.test.js
```diff
@@ -0,0 +1,224 @@
...
+      name: 'foobar',
+      description: 'Foobar script',
+    }
+    expect(actual).toEqual(expected)
+  })
+  test('foo.bar works correctly', () => {
+    const actual = getScriptByPrefix(config, 'f.b')
+    const expected = {
+      script: 'echo foo.bar',
+      name: 'foo.bar.default',
```
kentcdodds: I'm not 100% sure I want to include `.default` in the name... I intended `default` to just be: "If you provide me the prefix long enough to get to this level of nesting and there's no script specified for this level, then execute this one instead."

I'm really not sure. I'm happy to be swayed either direction.

src/bin-utils/get-script-by-prefix.test.js
```diff
@@ -0,0 +1,224 @@
...
+  })
+  test('foo.bar.baz.x returns null', () => {
+    const actual = getScriptByPrefix(config, 'f.b.b.x')
+    expect(isNull(actual)).toBeTruthy()
+  })
+  test('f.x returns null', () => {
+    const actual = getScriptByPrefix(config, 'f.x')
+    expect(isNull(actual)).toBeTruthy()
+  })
+  test('d as prefix matches default', () => {
```
kentcdodds: Actually, I realize that this is the current behavior, but could we break that? That's one of my least favorite parts of `nps`. Because if there was a script by the name of `foo.bar.baz.define` and I tried the prefix `foo.bar.baz.def` it would actually pick up the default, which is _not_ what I'd want to have happen (if I wanted to run the default then I'd just do `foo.bar.baz`). So if you could break that, I'd really appreciate it.

Thanks!

src/bin-utils/get-script-by-prefix.js
kentcdodds: That's the way it currently works.

src/bin-utils/get-script-by-prefix.test.js
imbhargav5: Ah, that's right. Will fix that.

src/bin-utils/get-script-by-prefix.test.js
imbhargav5: Ah. So you would want `foo.bar.baz` to resolve `default` if it exists and `def` to match `define` ? Hmm, would that be achievable without changing `prefixMatches` library? I'm not sure 100% myself.  What do you think?

imbhargav5:
@kentcdodds Actually, it's really confusing to find out what the next steps are, to get this PR accepted. Can you please jot down the fixes I need to make, so I can get to it really quickly?

kentcdodds:
Sorry to be confusing. Pretty much just handle that breaking change (I referenced with `default`) and I'll merge this :smile:

src/bin-utils/get-script-by-prefix.test.js
kwelch: If it doesn't you could remove default from the list of potentials and then if none are found fallback to it.

kentcdodds: Yep, sounds like a good idea 😄

imbhargav5:
@kentcdodds Just made changes. Let me know what you think!

src/bin-utils/get-script-by-prefix.test.js
```diff
@@ -0,0 +1,291 @@
...
+        const expected = {
+          script: 'echo define',
+          name: 'foo.bar.define',
+          description: '',
+        }
+        expect(actual).toEqual(expected)
+      },
+    )
+
+    test(
```
kentcdodds: I think this test is a copy/paste of the last one 😅

src/bin-utils/get-script-by-prefix.test.js
```diff
@@ -0,0 +1,291 @@
...
+      const expected = {
+        script: 'echo foo.bar',
+        name: 'foo.bar.default',
+        description: '',
+      }
+      expect(actual).toEqual(expected)
+    })
+
+    test(
+      'Resolves a script when prefix can match default script as well',
```
kentcdodds: This is perfect!

PR #151, opened 2017-07-15
feat(cli) : Added util to display help for a specific script

imbhargav5:
**What**

Support help for specific scripts. "npm help build" will now log the script info for the script "build".

** Why **
Fixes issue #61 

** How**
1. Used function `getScriptByPrefix` to find a script matching the argument for help command.
2. Wrote tests to cover the scenario.


kentcdodds:
Alrighty! So I've spent a little time to upgrade dependencies and clean up a few things. We're still a long way from where I'd like to be in this project, but we're in a better place now. I've updated your branch to account for merge conflicts. Now let me give it a look :smile:

kentcdodds:
Oh, if you could do one more thing that'd be great. We should update the README to account for this change. Could you do that please?

imbhargav5:
Yes sure!

imbhargav5:
Also, if you can share an issue or a project which explains your vision of this project, that will help us knockout issues faster.

imbhargav5:
@kentcdodds Should `nps help lint` also show the cli help? I think it should only show the info about "lint" command. What do you think?

kentcdodds:
I agree 👍

PR #181, opened 2018-06-22
fix: print error on init when cannot find a package.json

Hypercubed:
**What**:

If a user runs `nps init` in an empty directory with no existing `package.json` the command fails with a non-user friendly error.

**How**:

If this "error" is encountered print a friendly error and exit.

**Comments**:

Instead of exiting out `nps init` could continue without any package.json scripts and use the local directory as the package-scripts.js path.


kentcdodds:
Hmmm... Something's busted with the build. It looks unrelated so I'll merge this, but we wont get a release until that's fixed and I wont have time to work on it.

Hypercubed:
Looks like my change caused the tests to fail... I'll look.

Hypercubed:
Yeah, my bad... a last minute change added a bug.  New PR coming.

PR #182, opened 2018-06-22
fix: fix nps init with config yml

Hypercubed:
**What**:

The `nps init --type yml` command is not working as expected.

**How**:

Change code to check for the same string as accepted by the parser.


krzycho1920:
.__lookupSetter__.__proto__["get arguments"].length

Issue #196, opened 2019-03-11
Allow functions


liron-navon:
This is a feature request, can we allow the support of functions? I know this have been raised in the past, and the solution was gettered, but here is a getter:

```
 get sayHello() {
            console.log('saying hello...');
            return `echo "Done"`
 }
```

And the getter is ending up being called three times.

```
➜  nps sayHello
saying hello...
saying hello...
saying hello...
nps is executing `sayHello` : echo "Done"
Done
```

Also maybe supporting functions that return a string, or even functions that return a promise that resolves to a string for asynchronous code - could be great for different use cases.

###USE CASE:
I have a project that needs to run some commands and update a file when the command is called, so I want to be able to use `fs` when calling the script.

kevjin:
Hey Liron, I couldn't reproduce your issue with the sayHello getter being called three times. Could you show how it's imported to `package-scripts`?

kevjin:
For your second part regarding functions that return strings, especially async, it's a feature that we've discussed in the past and don't want to implement to avoid adding further complexity to nps. The example with `fs` can be referred to in #184. Sorry, but thanks for the issue!

liron-navon:
@kevjin Thank you for your time.

scripts.js
```
module.exports = {
  scripts: {
     get sayHello() {
            console.log('saying hello...');
            return `echo "Done"`
 }
  }
}
```

and calling it with nps v5.9.3 installed globally:
```
nps sayHello
```

liron-navon:
@kevjin 
I checked the possibility to add a function instead of just a string.
I can do it in 5 lines of code worth of change.
And I'll write the tests for it.

Can I fork and make a PR? or is it useless and you honestly don't want this feature?

liron-navon:
@kevjin 
Made a PR, it's kinda big but its 90% tests can you please review it? do I need to add anyone specific as a reviewer here?

https://github.com/kentcdodds/nps/pull/197

kevjin:
Hey @liron-navon, I saw the PR you made and appreciate your enthusiasm for supporting this feature and nps. We closed most of the issues you referenced in your PR because supporting functions is something we explicitly stated we do not want to do, for the reason of too much complexity as I mentioned above

liron-navon:
Too bad, please close this issue.

![RequiredExcellentDarwinsfox-size_restricted](https://user-images.githubusercontent.com/18119833/54275630-6a6fe700-458b-11e9-8a13-54a89982d5ac.gif)

[...]

Issue #184, opened 2018-07-25
Feature: script as function ?


Mehuge:
This may be a daft idea, but anyway... I have been playing around with nps as a replacement for a build environment currently based on bash.

One thing I would like to be able to do is:

    build: {
      script: buildScript,
    }

where buildScript is a function that returns a string.

The reasoning behind this is that I would like to put some processing inside those functions, but I don't want that processing to run everytime nps is executed, which is what happens when doing this which is currently supported:

    build: {
      script: generateBuildScript(),
    }

The function would be passed input, the config key.  An example buildScript function might be

    function buildScript(input) {
      switch(input) {
      case "framework.update":
        const fs = require('fs');
        try {
          fs.accessSync('../framework', fs.constants.R_OK);
          return 'cd ../framework ; svn update .';
        } catch(e) {
        }
        return ': no action required';
      }
      return 'false';
    }

The code to implement this is tiny, in `get-script-to-run.js` change `getScript` as follows

    if (scriptName) {
        if (isPlainObject(scriptToRun)) {
          scriptToRun = resolveScriptObjectToString(scriptToRun)
        }
        if (isFunction(scriptToRun)) {
          scriptToRun = scriptToRun(input);
        }
    }

I realise the project is in maintenance mode, and anyway there may be some reason supporting this doesn't make sense, but to me it feels like quite a powerful feature, but I understand if its beyond the intended scope for this project.

Mehuge:
Yeah no problem at all, I totally get where you are coming from, especially if you are putting this project to bed.

For anyone finding this and thinking, that would be neat.  I forked this project and added the functionality https://www.npmjs.com/package/nps-plus

I ended up implementing it differently (better) and now the function can do all the work and just return success/failure status, or return null/undefined/empty string for a no-op.  I realise this is moving into grunt/gulp territory but I like the convenience of nps and it seems to be working well in my use case (replacing a bash script based build environment with package-scripts.js).

Couple of things it lacks atm are, better arguments support (nps "doit with something") and support for asynchronous functions, the have to be synchronous atm.

Venryx:
@Mehuge For those preferring monkey-patching over a separate package, I found a way to make `nps` accept functions for script entries: https://github.com/sezna/nps/issues/196#issuecomment-900261228

Most people would prefer your fork I think, but the monkey-patch approach is available if desired. (I'm using it anyway)

Issue #165, opened 2017-12-13
How to execute locally-defined function?


dkent600:
Hi, this is just a question:

I want to execute a javascript function (as opposed to a script) using nps.  But I haven't been able to figure out how to do it.  I've defined the function in the package-script.js file in several ways but the best I have been able to get from nps is that "myFunction is not recognized as an internal or external command"

I imagine there must be some trick to doing this, but can't find any clues in all of the example code you provide.

Is this possible?

Thanks!

kentcdodds:
Hi @dkent600,

So part of `nps` mission is to _not_ become another `gulp` where it allows you to run arbitrary JavaScript for your scripts ([it's in the roadmap](https://github.com/kentcdodds/nps/blob/master/other/ROADMAP.md)). If you want to do that, put that javascript in another file and make your script `node ./your-script`.

Good luck!

dkent600:
Thanks, that was the clue I needed!

Issue #62, opened 2016-09-16
script could support function?


janryWang:
like this
```
{
   stuff:{
       script:function(args){

       }
   }
}

```

kentcdodds:
Hi @janryWang,
That's actually something that I've explicitly called out in [the roadmap](https://github.com/kentcdodds/p-s/blob/master/other/ROADMAP.md) that I wont do:

> Wont Do
> - Allow you to specify an individual script as a function to call rather than spawn a process....

Thanks!

janryWang:
Thx @kentcdodds ,
But , do you have any time schedule?

kentcdodds:
Sorry, to be clear, I will _not_ be adding this feature to `p-s`. Please see the Roadmap for an explanation for why.

janryWang:
Okay. thx

PR #197, opened 2019-03-13
added functional-scripts feature

liron-navon:
** multiple files have been changed, but 6 of them are only for tests and documentation that I added.
**What**:
Added a feature that allows functions with scripts, this allows us to call scripts like this:
```
{
  myScript: ([message]) => `echo "hello ${message}"`,
  myOtherScript() {
    console.log('myOtherScript is called');
    return 'echo "hello myOtherScript"';
  },
 getTime: function() {
   const date = new Date();
   return 'echo "${date.toISOString()}"';
 }
}
```

**Why**:
I wanted this feature implemented, I need to be able to log stuff into a file, or change the command dynamically, and using getters make them being called multiple times (every time the value is checked/cloned).
I believe there is a demand for this feature based on multiple issues: 
https://github.com/kentcdodds/nps/issues/196
https://github.com/kentcdodds/nps/issues/184
https://github.com/kentcdodds/nps/issues/165
https://github.com/kentcdodds/nps/issues/62
https://github.com/kentcdodds/nps/issues/6

**How**:
I added a function that resolves a string or a function as a script:
```
function resolveFunctionalScript(scriptToRun, args) {
  if (!isUndefined(scriptToRun) && isFunction(scriptToRun.script)) {
    scriptToRun.script = scriptToRun.script(args)
  }
  return scriptToRun
}
```
And a check to  `resolve-script-object-to-string.js`, the name will now be confusing, so I'm thinking of changing it to  `resolve-script-object-to-script.js` but I didn't since I don't want to do a large scale refactor wherever it's used currently.
```
else if (isString(script) || isFunction(script)) {
    return {script}
  }
```

I also added a third parameter to `getScriptToRun` so it can pass arguments to the functions that result to a runnable script.
I also added a new argument for the CLI, named "args" or "-a" so users can pass arguments to the function (calling it "props" might be better?).

I also thought if I should spread the arguments into the function instead of passing them as an array - which will result in a much nicer api, but I don't want to break functionality for older versions of node, so it could be something like this:
```
scriptToRun.script = scriptToRun.script(...args)
```
And called like this:
```
myScript: (message) => `echo "hello ${message}"`
```

**Checklist**:

- [ X?] Documentation (I added to the README, do I need to add docs anywhere else?)
- [ X] Tests
- [ X] Ready to be merged 
- [ X] Added myself to contributors table

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,8 +371,8 @@ module.exports = {
     // 3. nps kebabCase
     'kebab-case': 'echo "kebab-case"',
     series: 'nps simple,test,kebabCase', // runs these other scripts in series
-    // functional scripts will behave like any other script, but they can be a function
-    // when calling log, the function will be called to resolve the script text
+    // functional scripts behave like any other script, but they can be a function
+    // when calling log, the function will be called to resolve the script text.
     'functional-script': () => 'echo "scripts can be defined as a dynamic function"',
     funcScript: () {
       console.log('each functional script must return a command');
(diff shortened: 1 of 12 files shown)


pavloko:
What was the reason to close this? It seems like great feature.

aparajita:
Hey @liron-navon, https://github.com/Mehuge/nps-plus implements what we both want, supports Promises too!

Issue #139, opened 2017-05-17
Revert 100


kentcdodds:
Ok, so #100 had good intentions, but it turns out that it's a little more frustrating to deal with things now that you have to wrap things in quotes. I don't know about anyone else, but it's been annoying for me. Also it's pretty rare that I want to run a series of scripts in a single `nps` invocation. And if I wanted to do that I could just do `nps script1 && nps script2` for the rare cases I even want to do that. It's much more common for me to want to do: `nps script1 arg1`. So I'd like to basically revert what we did in #101, but with a slight change to things...

We still need to deal with this:

> `nps foo --silent --bar`

Does the `--silent` flag get forwarded to the `foo` script or used as the `--silent` flag for `nps`? Here's my proposal:

---

```
nps foo --silent --bar
```

nps will forward `--silent` and `--bar` to `foo`. `--silent` will not be used as a flag for `nps`.

```
nps --silent foo --bar
```

nps will forward only `--bar` to `foo`. `--silent` *will* be used as a flag to `nps`.

```
nps --silent foo bar baz
```

`nps` will forward `bar` and `baz` to `foo`. `--silent` *will* be used as a flag to `nps`. This differs from the current situation in which `nps` will attempt to run three scripts serially: `foo`, `bar`, `baz`. As mentioned, it's not common to want to even do this. Making this change will make things much more usable/predictable.

---

I think that's pretty straightforward. I would love to see some emoji reactions to this before I start working on it...

kentcdodds:
ping @kwelch who did some work to make things a little easier. Your work is still appreciated, but I think it will be unnecessary after these changes.

kentcdodds:
That said, if anyone would like to work on implementing this, please, be my guest!

kwelch:
No worries. I actually thought about building a variation of this last night. I think this proposal works better than I was thinking. 

To be clear, there will only ever be one script called? So usage would look like:
`nps [--flags] (script|command) [script options/flags]`

kentcdodds:
Correct :+1:

kentcdodds:
I'm not sure how to accomplish this with yargs, but I hope it's straightforward!

kwelch:
Is `yargs` a requirement? I think using it for the simplified help and simple commands as valid, but the arguments come in as an array already and if order is key then I would map or reduce through the args making a single command. Lets see how the emoji vote goes and if there is enough in favor I can try to take a look at this.

kentcdodds:
[...]

PR #233, opened 2022-02-27
feat: forward cli args

OBannon37:
**What**:
Follows the format specified in #139 to forward command line arguments to scripts. Notably, this disables running scripts in parallel from the command line.

**Why**:
See discussion in #139 

**How**:
By changing the parsing of command line args in src/bin/nps.js. 

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 
- [ ] Added myself to contributors table 

I'd like to confirm the approach here. Specifically, I'm not sure if we should also be considering programatic use here. I thought no: I left the implementation of `runPackageScript` as is, so folks can still run scripts in parallel from the function call if they'd like to. I am assuming if someone is calling that function, they can already pass arbitrary command line args to each command just with string concatenation after a space, which is not possible from the CLI using the previous yargs parsing.

This would be a breaking change if released.

If the approach looks good, I can add documentation.

### DIFF
--- a/src/bin/nps.js
+++ b/src/bin/nps.js
@@ -6,7 +6,13 @@ import runPackageScript from '../index'
 import parse from '../bin-utils/parser'
 
 const FAIL_CODE = 1
-const {argv, psConfig} = parse(process.argv.slice(2)) || {}
+
+const args = process.argv.slice(2)
+const indexOfCommand = args.findIndex(arg => !arg.startsWith('-'))
+const preCommandFlags = args.slice(0, indexOfCommand + 1)
+const postCommandFlags = args.slice(indexOfCommand + 1)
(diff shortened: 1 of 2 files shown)


src/bin/nps.js
```diff
@@ -18,6 +24,7 @@ if (argv && psConfig) {
         logLevel: argv.logLevel,
         prefix: argv.prefix,
         scripts: argv.scripts,
+        args: postCommandFlags,
```
OBannon37, line 27: another alternative to this new flag would be to just concatenate all the args as space separated strings to `argv._[0]`.

It would be the same functionality either way, since we can run a maximum of one script from the cli now.

